PR #109, opened 2019-03-19
slash-bedrock/share/common-code::198~216 : Working around fatal error

Kreyren:
CONCEPT
- Added codeblock from KULUS (https://github.com/RXT067/Scripts/blob/a2f78970e20f26af7cb75cf5489ab49b9b903328/KULUS/KULUS.sh#L51) that works around user requiring root permission. (KULUS was education project, not representable)
- Will be recoded in shell script, currently in bash.

Assuming finished mergable in bedrock Linux? if not close.

slash-bedrock/share/common-code::29~32 : Pulled by preferences set in text editor will redo.

Signed-off-by: Jacob Hrbek <werifgx@gmail.com>

### DIFF
--- a/src/slash-bedrock/share/common-code
+++ b/src/slash-bedrock/share/common-code
@@ -26,10 +26,10 @@ print_logo() {
 	#
 	# shellcheck disable=SC1117
 	cat <<EOF
-__          __             __      
-\ \_________\ \____________\ \___  
- \  _ \  _\ _  \  _\ __ \ __\   /  
-  \___/\__/\__/ \_\ \___/\__/\_\_\ 
+__          __             __
+\ \_________\ \____________\ \___
+ \  _ \  _\ _  \  _\ __ \ __\   /
+  \___/\__/\__/ \_\ \___/\__/\_\_\
 EOF
 	if [ -n "${1:-}" ]; then
 		printf "%35s\\n" "${1}"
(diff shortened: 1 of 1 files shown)


paradigm:
The idea you're proposing here, if I understand correctly, is to have `brl` detect when it's running as a non-root user but root is required.  Should that situation occur, prompt the user to re-run as root.

In principle it's not a bad idea, but I personally don't like it.  `brl` takes after an established convention for programs which sometimes need root and sometimes do not, for which users have workflows such as `sudo !!`.  For example, consider `apt install` (which typically requires root) and `apt list` (which typically does not).

If you want to continue to work on this for your own purposes, I don't think `require_root()` is the best place to do it.  Instead, I'd put it in `/bedrock/bin/brl`.  Split the case into those that need root and those that do not.  For those that do, you can wrap `su`/`sudo` around the `exec` line.  I'm actually strongly considering doing something like this with `strat` in an upcoming `brl` refactor for performance, rather than `strat`'ing repeatedly in a loop as it does in things like `brl status`.

Kreyren:
Noted, I understand that `require_root()` function may be used in `brl` which is supposed to run as non-root? if not provided codeblock would run only if `brl` specifically requires root permission instead of outputing err?

Can you elaborate?

paradigm:
> Noted, I understand that `require_root()` function may be used in `brl` which is supposed to run as non-root?

`require_root()` is a function which checks if the code it's running in is running as root and, if not, aborts with an appropriate error message.

It is used in various parts of Bedrock to ensure a proper error message is provided along code paths that require root if not run as root.  This includes `brl`, but also includes the installer/updater and the upcoming `pmm`.  It exists in `/bedrock/shared/common-code` because it is common to multiple parts of Bedrock.

Sometimes `brl` should run be run as root, and sometimes it should not.  It depends on what the user wants to do.  This pattern is established by programs such as `apt`, as I mentioned before.

> if not provided codeblock would run only if brl specifically requires root permission instead of outputing err?

If `require_root()` is not in place in a code path which requires root, a following operation which requires root will fail, likely with a less useful error message such as "Unexpected error occurred."

PR #114, opened 2019-04-19
Fix installer on systems with bsdtar

Skirmisher:
Some distros (e.g. [Adelie Linux](http://adelielinux.org/)) ship [bsdtar](https://github.com/libarchive/libarchive/wiki/ManPageBsdtar1), provided by libarchive, in place of GNU tar. The default behavior of bsdtar, when not provided the `f` flag with an argument, is to open the tape drive, `/dev/st0` on Linux. This is not what we want.

This patch explicitly specifies stdin as the input file for all tar invocations, and fixes the installer on distros with bsdtar.

### DIFF
--- a/src/installer/installer.sh
+++ b/src/installer/installer.sh
@@ -70,7 +70,7 @@ Please type \"Not reversible!\" without quotes at the prompt to continue:
 		abort "Warning not copied exactly."
 	fi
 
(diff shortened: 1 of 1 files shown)


paradigm:
Excellent, thanks!  If I touch a relevant subsystem - and I remember - I'll be sure to test against bsdtar in addition to GNU tar and busybox tar.

PR #119, opened 2019-05-30
brl-fetch: exherbo initial

Kreyren:
Created on demand, feedback and contribs welcommed.

```bash
	# ABSTRACT - Get exherbo bootstrap
	##X Download tarbar to $bootstrap_dir
	##X Extract tarbar in $bootstrap_dir/target_root because brl is then moving those to ${target_dir}
	##X Extract tarbar into /tmp so that you can flex on paradigm's shitty $bootstrap_dir solution
	## LOW_PRIO: Fetch different arch based on system used
	##X Configure paludis to use stratum specific /var/tmp/paludis directory to prevent conflicts with multiple paludis-based systems used
	### Get lib variable depending on libc used
	##X Fix permission for /var/tmp/paludis
	###X If directory is present remove it's content
	## Fix permission for /etc/paludis/repositories/installed.conf to install in builddir = /var/tmp/paludis differenciated from each stratum

```


Kreyren:
ah.. sorry wrong origin

PR #120, opened 2019-05-30
slash-bedrock/etc/bedrock.conf: removed share for /run

Kreyren:
Remove `/run` from global files since it's causing breakage on debian and exherbo.

To reproduce get openrc from gentoo and try to install `mariadb-server` on debian-buster with systemd and try to remove it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,12 +120,12 @@
 #
 #     make check
 
-VERSION=0.7.6
-CODENAME=Poki
+VERSION=9999
+CODENAME=Anubis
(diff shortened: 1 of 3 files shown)


paradigm:
- A global `/run` is needed for cross-stratum communication.  We can't simply remove it.  Whatever the issue is here, other solutions possible should be explored.
- This PR bleeds in all sorts of unrelated changes.  We're not renaming Bedrock to your fork.
- You're banned from the Bedrock community for reasons we've already discussed elsewhere.  Finding areas I haven't already explicitly locked you out of isn't helping your case.

PR #121, opened 2019-05-31
Add Exherbo support

Philantrop:
These commits add support for fetching Exherbo Linux (https://exherbo.org/). I'm Wulf "Philantrop" Krueger, a core distro developer for Exherbo (https://exherbo.org/developers.html) and would take care of this fetcher as a distro maintainer if you agree.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 bedrock-linux-*.sh
 build
 vendor
+*~
--- a/src/slash-bedrock/share/brl-fetch/distros/exherbo
+++ b/src/slash-bedrock/share/brl-fetch/distros/exherbo
@@ -48,10 +48,12 @@ list_releases() {
 }
 
 fetch() {
-        paludis_builddir=/var/paludis
+	# The following line will be changed to /var/tmp/paludis once /var/tmp
+	# is no longer global. For testing the github PR it temporarily stays.
+	paludis_builddir=/var/paludis
         paludis_repo_dir=/etc/paludis/repositories
-        paludis_user=paludisbuild
-        paludis_group=paludisbuild
+	paludis_user=paludisbuild
+	paludis_group=paludisbuild
 
 	ex_stage=exherbo-x86_64-pc-linux-gnu-current.tar.xz
 	ex_checksum=sha1sum
--- a/src/slash-bedrock/share/brl-fetch/distros/exherbo
+++ b/src/slash-bedrock/share/brl-fetch/distros/exherbo
@@ -60,12 +60,12 @@ fetch() {
 	mkdir -p "${ex_download}"
 
 	step "Downloading bootstrap software"
-	if [[ -f ${ex_download}/${ex_stage} ]]; then
+	if [[ -f "${ex_download}"/"${ex_stage}" ]]; then
 	    cp "${ex_download}/${ex_stage}" "${bootstrap_dir:-}"
 	    cp "${ex_download}/${ex_checksum}" "${bootstrap_dir:-}"
 	else
-	    wget ${target_mirror:-}/stages/"${ex_stage}" -O "${bootstrap_dir:-}"/"${ex_stage}"
-	    wget ${target_mirror:-}/stages/"${ex_checksum}" -O "${bootstrap_dir:-}"/"${ex_checksum}"
+	    wget "${target_mirror:-}"/stages/"${ex_stage}" -O "${bootstrap_dir:-}"/"${ex_stage}"
+	    wget "${target_mirror:-}"/stages/"${ex_checksum}" -O "${bootstrap_dir:-}"/"${ex_checksum}"
 	fi
 
 	step "Verifying stage tarball"
@@ -91,19 +91,19 @@ fetch() {
 
 	step "Creating Paludis' build-dir & fixing permissions"
 	mkdir -p "${target_dir:-}"/"${paludis_builddir}/"
-	chroot ${target_dir} chown -R ${paludis_user}:${paludis_group} "${paludis_builddir}"
-	chroot ${target_dir} chmod -R g+rwx "${paludis_builddir}"
+	chroot "${target_dir}" chown -R "${paludis_user}":"${paludis_group}" "${paludis_builddir}"
+	chroot "${target_dir}" chmod -R g+rwx "${paludis_builddir}"
 
 	step "Modifying ${paludis_user}"
-	chroot ${target_dir} usermod --home "${paludis_builddir}" ${paludis_user}
-	chroot ${target_dir} usermod -a -G $(stat -c %G /dev/tty) ${paludis_user}
+	chroot "${target_dir}" usermod --home "${paludis_builddir}" ${paludis_user}
+	chroot "${target_dir}" usermod -a -G "$(stat -c %G /dev/tty)" ${paludis_user}
 
 	case "${ex_stage}" in
 	     *-musl-*) ex_locale_conf=musl;;
 	     *) ex_locale_conf=true;;
 	esac
 
-	if [[ ${ex_locale_conf} == "true" ]]; then
+	if [[ "${ex_locale_conf}" == "true" ]]; then
 	    step "Preparing locales"
 
 	    # get local locales because they will bleed trough via the environment
@@ -113,21 +113,21 @@ fetch() {
 	    brl_loc=$(cfg_value "locale" "localegen" | cut -d ' ' -f1 -)
 	    ex_locales="${ex_locales} ${brl_loc}"
 
-	    for loc in ${ex_locales}; do
-	        ex_locale=${loc/\.*}
-	        ex_charset=${loc/*\.}
-		ex_charset=$(
+	    for loc in "${ex_locales}"; do
+	        ex_locale="${loc/\.*}"
+	        ex_charset="${loc/*\.}"
+		ex_charset="$(
 		    case "${ex_charset}" in
 		        *-*) echo ${ex_charset} | tr '[a-z]' '[A-Z]';;
 		        *) echo ${ex_charset/8/-8} | tr '[a-z]' '[A-Z]';;
 		    esac
-	        )
-	        chroot ${target_dir} localedef -i "${ex_locale}" -f "${ex_charset}" "${ex_locale}"."${ex_charset}"
+	        )"
+	        chroot "${target_dir}" localedef -i "${ex_locale}" -f "${ex_charset}" "${ex_locale}"."${ex_charset}"
 	    done
-	    echo LANG="$(cfg_value locale LANG)" > ${target_dir}/etc/env.d/99locale
-	elif [[ ${ex_locale_conf} == "musl" ]]; then
+	    echo LANG="$(cfg_value locale LANG)" > "${target_dir}"/etc/env.d/99locale
+	elif [[ "${ex_locale_conf}" == "musl" ]]; then
 	    step "No locale config for exherbo-musl (yet)"
 	fi
-	chroot ${target_dir} eclectic env update >/dev/null
+	chroot "${target_dir}" eclectic env update >/dev/null
 }
 
--- a/src/slash-bedrock/share/brl-fetch/distros/exherbo
+++ b/src/slash-bedrock/share/brl-fetch/distros/exherbo
@@ -53,20 +53,12 @@ fetch() {
         paludis_user=paludisbuild
         paludis_group=paludisbuild
 
-	ex_download=/var/tmp/brl-fetch
 	ex_stage=exherbo-x86_64-pc-linux-gnu-current.tar.xz
 	ex_checksum=sha1sum
 
-	mkdir -p "${ex_download}"
-
 	step "Downloading bootstrap software"
-	if [[ -f "${ex_download}"/"${ex_stage}" ]]; then
-	    cp "${ex_download}/${ex_stage}" "${bootstrap_dir:-}"
-	    cp "${ex_download}/${ex_checksum}" "${bootstrap_dir:-}"
-	else
-	    wget "${target_mirror:-}"/stages/"${ex_stage}" -O "${bootstrap_dir:-}"/"${ex_stage}"
-	    wget "${target_mirror:-}"/stages/"${ex_checksum}" -O "${bootstrap_dir:-}"/"${ex_checksum}"
-	fi
+	wget "${target_mirror:-}"/stages/"${ex_stage}" -O "${bootstrap_dir:-}"/"${ex_stage}"
+	wget "${target_mirror:-}"/stages/"${ex_checksum}" -O "${bootstrap_dir:-}"/"${ex_checksum}"
 
 	step "Verifying stage tarball"
 	cd "${bootstrap_dir:-}"
--- a/src/slash-bedrock/share/brl-fetch/distros/exherbo
+++ b/src/slash-bedrock/share/brl-fetch/distros/exherbo
@@ -91,15 +91,57 @@ fetch() {
 	step "Configuring"
 	setup_chroot "${target_dir:-}"
 
-	echo "builddir = \${root}${paludis_builddir}" >> "${target_dir}/${paludis_repo_dir}/arbor.conf"
-	echo "builddir = \${root}${paludis_builddir}" >> "${target_dir}/${paludis_repo_dir}/i686-pc-linux-gnu.conf"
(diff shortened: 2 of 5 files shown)


paradigm:
I'd absolutely love for you to take the role of Bedrock's Exherbo maintainer.  When I get the chance I'll update the project website indicating so, and work with you to get PR in.

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,147 @@
...
+	    if [[ -f ${ex_download}/${ex_stage} ]]; then
+	        cp "${ex_download}/${ex_stage}" "${bootstrap_dir:-}"
+		cp "${ex_download}/${ex_checksum}" "${bootstrap_dir:-}"
+	    else
+		wget ${target_mirror:-}/stages/"${ex_stage}" -O "${bootstrap_dir:-}"/"${ex_stage}"
+		wget ${target_mirror:-}/stages/"${ex_checksum}" -O "${bootstrap_dir:-}"/"${ex_checksum}"
+	    fi
+	)
+
+	step "Verifying stage tarball"
```
paradigm: This is great.  I need to add checksumming to other `brl fetch` back ends. :+1:

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,147 @@
...
+	    echo "builddir = \${root}${paludis_builddir}" >> "${target_dir}/${paludis_repo_dir}/arbor.conf"
+	    echo "builddir = \${root}${paludis_builddir}" >> "${target_dir}/${paludis_repo_dir}/i686-pc-linux-gnu.conf"
+	    echo "builddir = \${root}${paludis_builddir}" >> "${target_dir}/${paludis_repo_dir}/installed.conf"
+	    echo "builddir = \${root}${paludis_builddir}" >> "${target_dir}/${paludis_repo_dir}/../repository.template"
+	)
+
+	step "Creating Paludis' build-dir & fixing permissions"
+	(
+	    mkdir -p "${target_dir:-}"/"${paludis_builddir}/"
+	    chroot ${target_dir} chown -R ${paludis_user}:${paludis_group} "${paludis_builddir}"
```
paradigm: If `${target_dir}` has whitespace, this could expand problematically.  Consider:

```
sudo /bedrock/libexec/busybox sh
mkdir "foo bar"
cp /bedrock/libexec/busybox "foo bar/busybox"
x="foo bar"
chroot ${x} /busybox
```

prints

```
chroot: can't change root directory to 'foo': No such file or directory
```

Bedrock *tries* to enforce strata do not have whitespace in their names explicitly to fortify against issues such as this, but I prefer to take a layered approach rather than rely on it.  Thus, I'd prefer `${target_dir}` be quoted here.

Moreover, I've found the easiest way to ensure consistency on sometimes optional things is to just always mandate them.  I know you've defined `${paludis_user}` and `${paludis_group}` variables above without any white space or backslashes or anything funky and it doesn't make a difference, but I'd prefer the variable be quoted for consistency's sake.

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,147 @@
...
+	    chroot ${target_dir} chmod -R g+rwx "${paludis_builddir}"
+	)
+
+	step "Modifying ${paludis_user}"
+	(
+	    chroot ${target_dir} usermod --home "${paludis_builddir}" ${paludis_user}
+	    chroot ${target_dir} usermod -a -G $(stat -c %G /dev/tty) ${paludis_user}
```
paradigm: Just like with `${variables}`, I prefer `$(...)` to be quoted to ensure it's treated atomically.  Consider:

```
sudo /bedrock/libexec/busybox sh
mkdir "foo bar"
cp /bedrock/libexec/busybox "foo bar/busybox"
chroot $(echo "foo bar") /busybox
```

printing

```
chroot: can't change root directory to 'foo': No such file or directory
```

Even if there was some guarantee that it's not strictly needed in this instance, I'd prefer to just quote `$(...)` everywhere for consistency's sake.

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,147 @@
...
+	)
+
+	case "${ex_stage}" in
+	     *-musl-*) ex_locale_conf=musl;;
+	     *) ex_locale_conf=true;;
+	esac
+
+	if [[ ${ex_locale_conf} == "true" ]]; then
```
paradigm: There are weird scenarios where Bedrock's code is sometimes run by shell other than `/bedrock/libexec/busybox`.  The most obvious case is when hijacking.  Thus, the preference is towards more portable code, particularly when such an option presents itself with no functional difference.  Thus, I'd prefer

```
if [ "${ex_locale_conf}" = "true" ]; then
```

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,147 @@
...
+	    step "Preparing locales"
+
+	    # get local locales because they will bleed trough via the environment
```
paradigm: `through` ;)

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,147 @@
...
+	    ex_locales=$(locale | sed -e "s:.*=::" | sort -u)
```
paradigm: `brl fetch` tries to rely only on what Bedrock provides (which is mostly just busybox) ensure it works irrelevant of whatever other strata.  We're not providing a `locale` command.  We may need some other strategy for what you're attempting here.

I am not as knowledgable about locale details as I should be, and I could be missing something.  I think you're trying to get the `LANG` variable value, which Bedrock collects in `bedrock.conf` for this purpose.  I _think_ you can achieve what you're after here with

```
ex_locales=$(cfg_value "locale" "LANG")
```

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,147 @@
...
+
+	    # get the locale from bedrock.conf
+	    brl_loc=$(cfg_value "locale" "localegen" | cut -d ' ' -f1 -)
```
paradigm: `[locale]/localegen` can have multiple, comma separated values.  Bedrock provides a `cfg_values` (with an `s`) that splits the commas into separate lines for you.  You might want to use that here and loop over the lines.

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,147 @@
...
+	echo "rolling"
+}
+
+list_releases() {
+	echo "rolling"
+}
+
+fetch() {
+        paludis_builddir=/var/paludis
+        paludis_repo_dir=/etc/paludis/repositories
```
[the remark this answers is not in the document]
Philantrop: paludis_repo_dir points to Exherbo's standard directory, yes. It's just there because I try to avoid verbatim strings among code. 

paludis_builddir is another beast: That's the shared /var/tmp thing we discussed on IRC as the standard directory would indeed be /var/tmp/paludis. 

Sharing /var/tmp could be problematic for similar reasons for at least Gentoo and Funtoo as well, AFAIK. So, making it local would be good, I'd say.

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,147 @@
...
+
+	step "Downloading bootstrap software"
+	(
+	    if [[ -f ${ex_download}/${ex_stage} ]]; then
+	        cp "${ex_download}/${ex_stage}" "${bootstrap_dir:-}"
+		cp "${ex_download}/${ex_checksum}" "${bootstrap_dir:-}"
+	    else
+		wget ${target_mirror:-}/stages/"${ex_stage}" -O "${bootstrap_dir:-}"/"${ex_stage}"
+		wget ${target_mirror:-}/stages/"${ex_checksum}" -O "${bootstrap_dir:-}"/"${ex_checksum}"
+	    fi
```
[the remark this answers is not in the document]
Philantrop: Yes, that would be useful as a global feature. 

I'd like to ask for an exception here, though, since we simply don't have mirrors for our stages. Everyone downloading a stage will always hit our main development machine.
I'll gladly replace it with a well-designed alternative but for the moment I'd appreciate it if we could leave it in.

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,147 @@
...
+        paludis_group=paludisbuild
+
+	ex_download=/var/tmp/brl-fetch
+	ex_stage=exherbo-x86_64-pc-linux-gnu-current.tar.xz
+	ex_checksum=sha1sum
+
+	mkdir -p "${ex_download}"
+
+	step "Downloading bootstrap software"
+	(
```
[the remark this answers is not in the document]
Philantrop: Completely agreed. I was trying to mimic the style. I'm removing the useless parenthesis.

src/slash-bedrock/share/brl-fetch/distros/exherbo
Philantrop: Thanks. Here, too, I think, some global framework would be nice. I know other distros offer the possibility as well.

src/slash-bedrock/share/brl-fetch/distros/exherbo
Philantrop: I agree. Path name always should be quoted. I simply messed up there. :)

I'm going to quote paludis_user and *_group as well, albeit POSIX says those simply must not contain whitespace or similar for portability's sake.

src/slash-bedrock/share/brl-fetch/distros/exherbo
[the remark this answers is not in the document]
paradigm: Lets change this PR to make use the default `/var/tmp/` (or `/var/tmp/paludis`?) with the expectation that `/var/tmp` will be removed from the global list before this is merged.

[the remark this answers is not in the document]
Philantrop: Yes, that would be best. I'll do that. And it's in fact /var/tmp/paludis. :-)

src/slash-bedrock/share/brl-fetch/distros/exherbo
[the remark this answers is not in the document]
paradigm: Ah, I was modelling this for debugging/injecting, not caching.  Minimizing mirror usage is completely reasonable.  I *really* don't want Bedrock getting a bad reputation for abusing mirrors.

I'd still prefer a some discussion on how to implement it with thought towards a generalized `brl fetch` solution before merging any `brl fetch` back end with such a feature.  Instead of removing caching from this MR, we can just work on the generalized solution first, provided you don't mind waiting a bit longer for this to merge.

Here's how I'm imagining a generalized `brl fetch` caching system right now:

- `/bedrock/libexec/brl-fetch` sets some variable - I'm currently thinking `${cache}` but am up for suggestion - that each back end inherits and uses.
- `${cache}` expands to `/bedrock/var/cache/brl-fetch/${distro}-${target_release}-${target_arch}`.  While each back-end *could* handle release/arch concerns, I don't think it's worth the complexity; easier to just have `brl fetch` segregate them.
- `brl fetch` will flock the cache directory.  If another `brl fetch` instance is grabbing the same release of the same distro for the same architecture at the same time, the new one errors out.  While it should definitely be possible to handle multiple simultaneous instances, I don't think it's worth the complexity.  Moreover, with caching, sequential access should end up being faster.
- `bedrock.conf` will contain some setting for how long the `brl fetch` cache should be kept around.  Calls to `brl fetch` after this time period will clear the cache.  Off the cuff I'm thinking of calling the section `[brl-fetch-cache]`, the key `expiration`, adding a comment explaining the unit is in days, defaulting this to `90`, but I'm definitely up for suggestion.
- The cache directory's timestamp is updated every `brl fetch` for the given distro/release/arch.  If the cache doesn't change, it sticks around.  This could be useful for distros like Debian that could re-use the same cached value for potentially years.
- Back ends are expected to check if the cached value is out of date every `brl fetch`.  If there are things like checksums, they should be re-downloaded every `brl fetch`.

I have not spent a lot of time thinking this through.  Feel free to push back on any detail, raise counter proposals, etc.

If this ends up being the system, I can probably add the `bedrock.conf` setting and shared caching code to `brl fetch` reasonably quickly.  May take me a bit to implement for all the other back ends, but they don't have to all get it in the same release.  I think the only changes you would need here are:

- Using the inherited `${cache}` (or rather `${cache:-}` to make `make check` happy, as it does not know the variable will be inherited).
- Changing the code to download the checksum every time, clear the old cached value if it is out of date or corrupt, then (re)download the new item.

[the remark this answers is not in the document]
Philantrop: Sounds good to me.

Philantrop:
I hope I've addressed everything. :-)

paradigm:
Looks like you did not:

- Change `paludis_builddir=/var/paludis` to `paludis_builddir=/var/tmp/paludis`
- Use the proposed `${cache}` system.

This makes sense, as those are both dependent on changes elsewhere, and would result in the code not working in its existing context.  Provided we don't forget to make those changes before merging, everything looks good to me.  I'll get back to you once the changes needed elsewhere are made at which time we can make the two above mentioned changes, rebase, and merge.  Provided nothing comes up that requires rushing an update out, my guess is this will be in 0.7.7 released in about two weeks.

nift4:
@Philantrop @paradigm said this PR will be in 0.7.8, not 0.7.7 like said above.

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,94 @@
...
+	echo "builddir = \${root}${paludis_builddir}" >> "${target_dir}/${paludis_repo_dir}/../repository.template"
+
+	step "Creating Paludis' build-dir & fixing permissions"
+	mkdir -p "${target_dir:-}"/"${paludis_builddir}/"
+	chroot "${target_dir}" chown -R "${paludis_user}":"${paludis_group}" "${paludis_builddir}"
+	chroot "${target_dir}" chmod -R g+rwx "${paludis_builddir}"
+
+	step "Modifying ${paludis_user}"
+	chroot "${target_dir}" usermod --home "${paludis_builddir}" ${paludis_user}
+	chroot "${target_dir}" usermod -a -G "$(stat -c %G /dev/tty)" ${paludis_user}
```
paradigm: Setting this dynamically is great <3

src/slash-bedrock/share/brl-fetch/distros/exherbo
```diff
@@ -0,0 +1,94 @@
...
+
+	step "Configuring Paludis' repositories"
+	echo "builddir = \${root}${paludis_builddir}" >> "${target_dir}/${paludis_repo_dir}/arbor.conf"
+	echo "builddir = \${root}${paludis_builddir}" >> "${target_dir}/${paludis_repo_dir}/i686-pc-linux-gnu.conf"
+	echo "builddir = \${root}${paludis_builddir}" >> "${target_dir}/${paludis_repo_dir}/installed.conf"
+	echo "builddir = \${root}${paludis_builddir}" >> "${target_dir}/${paludis_repo_dir}/../repository.template"
+
+	step "Creating Paludis' build-dir & fixing permissions"
+	mkdir -p "${target_dir:-}"/"${paludis_builddir}/"
+	chroot "${target_dir}" chown -R "${paludis_user}":"${paludis_group}" "${paludis_builddir}"
```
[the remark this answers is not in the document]
paradigm: Actually scratch that, I figured out how to get `shellcheck` to understand this situation a touch better.  Once I push caching there should be plenty of examples to reference.  Given this, we'll go the other way and drop all the `:-` we can so `shellcheck` can catch typo'd variable names.

.gitignore
```diff
@@ -1,3 +1,4 @@
 bedrock-linux-*.sh
 build
 vendor
+*~
```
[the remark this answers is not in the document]
Philantrop: Oops, that wasn't even supposed to be included. Will be gone with the next iteration.

Philantrop:
Closed in favour of PR #138

PR #130, opened 2019-07-29
[0.7.8] Add LVM2 support everywhere

### LINKED ISSUE #129: LVM support
I know this is listed on the website. Here I want to discuss solutions.

Is http://www.tldp.org/HOWTO/LVM-HOWTO/lvm2_boot.html enough to make lvm work?

nift4:
# Current concept
* Build LVM2 statically without udev and selinux support
* Run the binaries in /sbin/init before `mount -a`
# Current progress
- [X] Download LVM2 in Makefile
- [X] Get it to build working bins
- [X] Integrate it into /sbin/init
# Current status
- Needs testing by @paradigm and is getting (hopefully) merged then
- Hopefully in Bedrock Linux 0.7.8(beta1)
# Tested with...
... Debian 10 (systemd) sda1=/ sda2=sda2_crypt sda2_crypt=lvmpv lvmlv1=/home lvmlv2=swap
... Void Linux musl (runit) No LVM
... Void Linux musl (runit) /mnt/test LVM
... Gentoo (openrc) with LVM with multiple volumes by @runningnak3d
#### Home LVM by @runningnak3d
... Gentoo (openrc)
... Void Linux (runit)
... Ubuntu (systemd)
... Fedora (systemd)
... Artix (openrc)
... Artix (runit)
... Devuan (openrc)
... Arch (systemd)
# Other things
https://github.com/bedrocklinuxcommunity/bedrocklinux-userland is where the development happens.
This PR fixes #129


.gitignore
```diff
@@ -1,3 +1,4 @@
 bedrock-linux-*.sh
 build
 vendor
+.*.swp
```
paradigm: I prefer not to add non-specific-codebase-related things like this to `.gitignore`.  Happily, git provides alternatives which you may find suitable.

For user-specific things which are also repo-specific - if you only generate `*~` files when working on Bedrock - git offers `$GIT_DIR/info/exclude`.  From `man 5 gitignore`:

> Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are specific to one user’s workflow) should go into the $GIT_DIR/info/exclude file.

If you create `*~` files all over the place and you want git to ignore them everywhere, you can use `core.excludeFile` (which defaults to `~/.config/git/ignore`):

> Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

Makefile
```diff
@@ -140,6 +141,9 @@ INDENT_FLAGS=--linux-style --dont-line-up-parentheses \
 	--continuation-indentation8 --indent-label0 --case-indentation0
 WERROR_FLAGS=-Werror -Wall -Wextra -std=c99 -pedantic
 
+LIBAIOVERSION=0.3.112
+LVMVERSION=2.03.05
+
```
paradigm: Where possible, I prefer to have the build system automatically determine the latest stable version of a given upstream project to save some maintenance work.  If an upstream update breaks something, that's okay - it's a canary to let us know to update the build system.  We can always manually pull an older version if need be.

You're following this pattern already with `util-linux`.  Why not libaio or lvm?  We can certainly make exceptions if there's good reason.  Whatever the reason is, we should probably add a comment.

Makefile
```diff
@@ -366,6 +373,46 @@ $(COMPLETED)/uthash: vendor/uthash/.success_fetching_source $(COMPLETED)/builddi
 	touch $(COMPLETED)/uthash
 uthash: $(COMPLETED)/uthash
 
+vendor/libaio/.success_retrieving_source:
+	rm -rdf vendor/libaio/
```
paradigm: Does the `-d` flag do anything given the `-r` flag is already in use?  If not, mind changing the various `rm -rdf`'s to `rm -rf` for consistency with the rest of the file?  It's super minor but I find consistency with this kind of thing helps people new to reading a file like this who would otherwise wonder why one thing is used in one place and another is in another.

Makefile
```diff
@@ -366,6 +373,46 @@ $(COMPLETED)/uthash: vendor/uthash/.success_fetching_source $(COMPLETED)/builddi
...
+	mkdir vendor/libaio
+	wget -O- http://ftp.de.debian.org/debian/pool/main/liba/libaio/libaio_$(LIBAIOVERSION).orig.tar.xz | tar xJ -C vendor/libaio -f -
+	mv vendor/libaio/libaio-$(LIBAIOVERSION)/* vendor/libaio
+	rm -rdf vendor/libaio/libaio-$(LIBAIOVERSION)
```
paradigm: This directory *should* be empty at this point, right?  We can `rmdir` here instead of `rm -rdf`.  The advantage of doing this is that, if somehow our expectation is wrong, we'll get a well placed error message, which can save on debugging efforts.

.gitignore
nift4: Finished, altough you copied this comment from Exherbo PR and wrote about `*~` instead of `.*.swp`

Makefile
nift4: I did not found an git mirror.

Makefile
nift4: No. `rm -d` equals `rmdir`.

Makefile
nift4: I think there is a .gitignore in the tarball and because of that it failed. Should I copy dotfiles or delete them?

nift4:
Short info: The libaio rm comment is markes as outdated because there was a `rm -rdf`.

Makefile
nift4: LibAIO does not even have a website, shows 404.

nift4: Okay found pagure.io/libaio, seems active.

nift4: Solved for libaio.

paradigm: Poking around at various distro webpages what they consider the libaio home page, I've found:

- [The Debian packages page for libaio](https://packages.debian.org/source/sid/libaio) lists https://pagure.io/libaio
- [The Ubuntu packages page for libaio](https://packages.ubuntu.com/eoan/libaio-dev) lists https://pagure.io/libaio
- [The Gentoo Linux page for libaio](https://packages.gentoo.org/packages/dev-libs/libaio) lists https://pagure.io/libaio
- [The Arch Linux packages page for libaio](https://www.archlinux.org/packages/core/x86_64/libaio/) lists http://lse.sourceforge.net/io/aio.html however the [PKGBUILD grabs from Debian](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/libaio)
- [The Alpine Linux packages page for libaio](https://pkgs.alpinelinux.org/package/edge/main/x86/libaio) lists http://lse.sourceforge.net/io/aio.html however the [APKGBUILD grabs from Debian](https://git.alpinelinux.org/aports/tree/main/libaio/APKBUILD)
- [The Void Linux srcpkg for libaio](https://github.com/void-linux/void-packages/blob/master/srcpkgs/libaio/template) lists http://lse.sourceforge.net/io/aio.html but grabs the source from Debian.

Given everyone who does not indicate upstream page is https://pagure.io/libaio is following Debian, and everyone else - including Debian - uses https://pagure.io/libaio, I'm inclined to think https://pagure.io/libaio is our best bet here.  https://pagure.io/libaio does have a tagged git repo we can use.  I _think_ pagure.io is affiliated with Fedora, so this makes some sense.

Provided you agree with my reasoning here, lets use this git repo for libaio: https://pagure.io/libaio.git

Doing the same dance for LVM2, I found:

- [Debian package page](https://packages.debian.org/sid/lvm2) lists http://sources.redhat.com/lvm2/
- [Ubuntu package page](https://packages.ubuntu.com/eoan/lvm2) lists http://sources.redhat.com/lvm2/
- [Gentoo Linux page](https://packages.gentoo.org/packages/sys-fs/lvm2) lists https://sourceware.org/lvm2/ which forwards to https://sourceware.org/lvm2/
- [Arch Linux page](https://www.archlinux.org/packages/core/x86_64/lvm2/) lists https://sourceware.org/lvm2/ which forwards to https://sourceware.org/lvm2/
- [Alpine Linux page](https://pkgs.alpinelinux.org/package/edge/main/x86/lvm2) lists https://sourceware.org/lvm2/ which forwards to https://sourceware.org/lvm2/
- [The Void Linux srcpkg](https://github.com/void-linux/void-packages/blob/master/srcpkgs/lvm2/template) lists https://sourceware.org/lvm2/ which forwards to https://sourceware.org/lvm2/

Hoping over to https://sourceware.org/lvm2/ I found:

> The source code is stored in git and can be accessed with
> git clone git://sourceware.org/git/lvm2.git
> 
> There is a web interface to git at http://sourceware.org/git/?p=lvm2.git 

At http://sourceware.org/git/?p=lvm2.git  I also see

> git://sourceware.org/git/lvm2.git
> ssh://sourceware.org/git/lvm2.git
> http://sourceware.org/git/lvm2.git

I've found https://sourceware.org/git/lvm2.git also works, which is slightly preferable here so no one MITM's the upstream source acquisition

Provided you agree with my reasoning here, lets use this git repo for lvm2: https://sourceware.org/git/lvm2.git

Makefile
paradigm: Ahh, okay.  That does explain it.  In that case I the `rm -r` is fine.  Apologies, I probably should have checked instead of assuming here.

Given the discussion about using git we're having elsewhere, this'll likely be a non-issue either way.

Makefile
nift4: Sadly the pagure repo didn't worked in my q&d test. Was it my copy-paste intelligence and/or it is working for you?

nift4: Ah okay, the grep fails. Hmm...

nift4: To be honest: I can't help here because I have no idea what that piece of code does.

nift4: Also LVM2 fails. Seems the tag format is invalid. I am going to resolve thus and open another thread.

nift4:
I am going to self-review my code to look for everything but we have to look at this first.

Makefile
paradigm: > To be honest: I can't help here because I have no idea what that piece of code does.

Assuming you mean these blocks:

```
	git clone --depth=1 \
		-b `git ls-remote --tags 'https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git' | \
		awk -F/ '{print $$NF}' | \
		sed 's/^v//g' | \
		grep '^[0-9.]*$$' | \
		sort -t . -k1,1n -k2,2n -k3,3n -k4,4n -k5,5n | \
		tail -n1 | \
		sed 's/^/v/'` 'https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git' \
		vendor/util-linux
```

They ask git to get the source code of the latest stable release.  I agree they're not the most readable; I need to find a better way to write them.  Just copy/pasting those blocks won't necessarily work because people use different formats for their tags.  You may have just gotten lucky with util-linux.

The key parts are:

```
git clone --depth 1 -b <branch> <url> <location>
```

which grabs the commit (and only that commit, not the full history) of the specified branch of the specified URL and places it in the specified location and

```
git ls-remote --tags <url>
```

which lists the remote tags.  Everything else - the grep, sed, awk, etc - works on the list of tags to find what appears to be the latest stable one.

Looks like libaio prefixes `libaio.` to their tags.  For example, `libaio-0.3.111` rather than `0.3.111`.  Try this:

```
	git clone --depth=1 \
		-b `git ls-remote --tags 'https://pagure.io/libaio.git' | \
		awk -F/ '{print $$NF}' | \
		sed 's/^libaio-//g' | \
		grep '^[0-9.]*$$' | \
		sort -t . -k1,1n -k2,2n -k3,3n -k4,4n -k5,5n | \
		tail -n1 | \
		sed 's/^/libaio-/'` 'https://pagure.io/libaio.git' \
		vendor/libaio
```

Looks like LVM2 tags use a `_` instead of a `.` in the tag name which breaks the logic used with util-linux.  For example, `v2_03_05` instead of `v2.03.05`.  Try this:

```
	git clone --depth=1 \
		-b `git ls-remote --tags 'https://sourceware.org/git/lvm2.git' | \
		awk -F/ '{print $$NF}' | \
		sed -e 's/^v//g' -e 's/_/./g' | \
		grep '^[0-9.]*$$' | \
		sort -t . -k1,1n -k2,2n -k3,3n -k4,4n -k5,5n | \
		tail -n1 | \
		sed -e 's/^/v/' -e 's/[.]/_/g'` 'https://sourceware.org/git/lvm2.git' \
		vendor/lvm2
```

.gitignore
paradigm: That I did >.<

nift4:
I unserstand what git does, my problems are the regular expressions.

nift4:
Both snippets failed at the grep.

paradigm:
Did you run them in a shell outside of make?  They both use make-specific double `$`.  Change all instances of `$$` to `$` if you want to run them directly.

The tag lookup logic works fine for me in both instances.  If it's failing for you I'll need more details to understand where our setups differ, such as the specific error message.

I did find that sourceware.org's git server doesn't seem to understand `--depth=1`.  You'll have to remove that, apparently.

If you want to try outside of make, such as in a termux shell, try:

```
	git clone --depth=1 \
		-b `git ls-remote --tags 'https://pagure.io/libaio.git' | \
		awk -F/ '{print $NF}' | \
		sed 's/^libaio-//g' | \
		grep '^[0-9.]*$' | \
		sort -t . -k1,1n -k2,2n -k3,3n -k4,4n -k5,5n | \
		tail -n1 | \
		sed 's/^/libaio-/'` 'https://pagure.io/libaio.git' \
		vendor/libaio
```

```
	git clone \
		-b `git ls-remote --tags 'https://sourceware.org/git/lvm2.git' | \
		awk -F/ '{print $NF}' | \
		sed -e 's/^v//g' -e 's/_/./g' | \
		grep '^[0-9.]*$' | \
		sort -t . -k1,1n -k2,2n -k3,3n -k4,4n -k5,5n | \
		tail -n1 | \
		sed -e 's/^/v/' -e 's/[.]/_/g'` 'https://sourceware.org/git/lvm2.git' \
		vendor/lvm2
```

When you enter it into the Makefile, double the `$` up:

```
	git clone --depth=1 \
		-b `git ls-remote --tags 'https://pagure.io/libaio.git' | \
		awk -F/ '{print $$NF}' | \
		sed 's/^libaio-//g' | \
		grep '^[0-9.]*$$' | \
		sort -t . -k1,1n -k2,2n -k3,3n -k4,4n -k5,5n | \
		tail -n1 | \
		sed 's/^/libaio-/'` 'https://pagure.io/libaio.git' \
		vendor/libaio
```

```
	git clone \
		-b `git ls-remote --tags 'https://sourceware.org/git/lvm2.git' | \
		awk -F/ '{print $$NF}' | \
		sed -e 's/^v//g' -e 's/_/./g' | \
		grep '^[0-9.]*$$' | \
		sort -t . -k1,1n -k2,2n -k3,3n -k4,4n -k5,5n | \
		tail -n1 | \
		sed -e 's/^/v/' -e 's/[.]/_/g'` 'https://sourceware.org/git/lvm2.git' \
		vendor/lvm2
```

nift4:
I commited my tested work :)

nift4:
You can do this while merging. Because I pushed the commits I can not rebase it without deleting the branch.

nift4:
If you did not knew about this GitHub feature: https://help.github.com/en/articles/merging-a-pull-request#merging-a-pull-request-on-github

nift4:
Ah, --force-push is something else. I didn't knew about this Git feature :joy:

nift4:
Okay I'll try it.

nift4:
Voilà, I got it.

nift4:
Okay.

nift4:
@runningnak3d any testing progress?

nift4:
@runningnak3d Which init you used in Devuan?

nift4:
@runningnak3d Again big thanks for testing!

nift4:
@runningnak3d Init? (I ask because we have to override things in systemd and don't know about other inits)

nift4:
SysVInit is not supported by Bedrock Linux itself @runningnak3d

nift4:
@runningnak3d If you don't mind I would prefer if you test the really freaking out setup on Ubuntu because it has the most chaotic init I know (Upstart-Systemd-SystemV mix) (And it's my favorite distro hehe)

nift4:
@runningnak3d https://bedrocklinux.org/0.7/compatibility-and-workarounds.html in the table

nift4:
@paradigm Fixed merge conflicts.

nift4:
@paradigm What's your progress about testing? [I want it merged before it creates more conflicts.]

nift4:
Finished, altough the merge conflict is here again because I had to reset the branch after being to silly to rebase an merge commit without triggering all the changes to merge. @paradigm

nift4:
So @paradigm I WILL NOT rebase the merge commit because I can't get it right (It wants to merge all your changes since 0.7.6). Please do it via the GitHub UI.

paradigm:
No worries about rebasing; I understand how that can be difficult.

There were still some minor items such as a missing `-p` flag on mkdir, but I'll clean those up myself in another commit rather than continue to delay this PR.

There's a number of other things I'd like to get merged in before cutting 0.7.8beta1.  My guess is it'll be out early next week.

paradigm:
Many thanks for your efforts here.  This was something Bedrock greatly needed.

PR #137, opened 2019-08-31
installer: Add message about updating in sanity check

nift4:
@paradigm Is it right that we can offline-update via `brl update` if we give it a script? It's not documented (At least not in commands on the website)

### DIFF
--- a/src/installer/installer.sh
+++ b/src/installer/installer.sh
@@ -113,7 +113,7 @@ Please type \"Not reversible!\" without quotes at the prompt to continue:
 	elif [ -e /bedrock/ ]; then
 		# Prefer this check at end of sanity check list so other sanity
 		# checks can be tested directly on a Bedrock system.
-		abort "/bedrock found.  Cannot hijack Bedrock Linux."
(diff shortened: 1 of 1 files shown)


paradigm:
I don't see any real benefit to this PR.  I've never heard of anyone who tried to update a Bedrock system by running the script with `--hijack` and ended up confused.  I _have_ run into people who try to hijack a Bedrock system to see what happens, which is why the check is as it is now.

I do see potential downsides:

- This adds more noise for people who read the code base
- This adds more noise for people who try to hijack a Bedrock install
- This risks confusing people who try to hijack a pre-0.7 Bedrock install to update it to 0.7.

You're welcome to make an argument in favor of this PR if you think I'm missing something.

> @paradigm Is it right that we can offline-update via brl update if we give it a script? It's not documented (At least not in commands on the website)

It's documented in `brl update --help` if you're on an update new enough to support the feature.

The website documentation is a bit behind.  Catching it up is on my to-do list.  Right now I'm focused on more pressing things.  I have a _huge_ backlog.

PR #143, opened 2019-09-06
exherbo{,-musl}: Actually create paludisbuild (group & user)

Philantrop:


paradigm:
Ah, I see.  The fix I proposed [here](https://github.com/bedrocklinux/bedrocklinux-userland/commit/4827a27be4cd7e07369e19936f96504cf9f4f267) ended up breaking things, as you can't `addgroup` until you've confirmed the user exists.  Mea culpa.

Instead of doing this, I'd like to see if we can fix it another way, if you don't mind.

`brl fetch` _used_ to add users/groups in each per-fetch back end, which became something of a pain to maintain.  I generalized it to [call a function](https://github.com/bedrocklinux/bedrocklinux-userland/blob/6caca6e91aba92c4bba16ce65fb7e0357089277d/src/slash-bedrock/libexec/brl-fetch#L1133) just after [running the per-back-end `fetch` function](https://github.com/bedrocklinux/bedrocklinux-userland/blob/6caca6e91aba92c4bba16ce65fb7e0357089277d/src/slash-bedrock/libexec/brl-fetch#L1113) which [imports any new users and groups it sees with the newly fetched stratum](https://github.com/bedrocklinux/bedrocklinux-userland/blob/6caca6e91aba92c4bba16ce65fb7e0357089277d/src/slash-bedrock/libexec/brl-fetch#L405-L451).

The issue here wasn't your original `usermod` within the chroot - that was fine.  The issue was `import_users_and_groups()` only added users and groups, but didn't add users _to_ groups.

What I'd like to do is revert back to your original `chroot "${target_dir}" usermod ...` strategy and expand `import_users_and_groups()` to add users _to_ groups.

Something like (untested, off the top of my head):

```
while read -r line; do
	if [ -z "$(echo "${line}" | awk -F: '{print$4}') ]; then
		continue
	fi
	group="$(echo "${line}" | awk -F: '{print$1}')"
	for user in $(echo "${line}" | cut -d: -f4- | sed 's/,/ /g'); do
		addgroup "${user}" "${group}"
	done
done <"${group_file}"
```

I can go after that if you'd like, or you're welcome to take a crack at it if you prefer; no strong preference here.

Philantrop:
> Ah, I see. The fix I proposed [here](https://github.com/bedrocklinux/bedrocklinux-userland/commit/4827a27be4cd7e07369e19936f96504cf9f4f267) ended up breaking things, as you can't `addgroup` until you've confirmed the user exists. Mea culpa.
> 
> Instead of doing this, I'd like to see if we can fix it another way, if you don't mind.
> 
> `brl fetch` _used_ to add users/groups in each per-fetch back end, which became something of a pain to maintain. I generalized it to [call a function](https://github.com/bedrocklinux/bedrocklinux-userland/blob/6caca6e91aba92c4bba16ce65fb7e0357089277d/src/slash-bedrock/libexec/brl-fetch#L1133) just after [running the per-back-end `fetch` function](https://github.com/bedrocklinux/bedrocklinux-userland/blob/6caca6e91aba92c4bba16ce65fb7e0357089277d/src/slash-bedrock/libexec/brl-fetch#L1113) which [imports any new users and groups it sees with the newly fetched stratum](https://github.com/bedrocklinux/bedrocklinux-userland/blob/6caca6e91aba92c4bba16ce65fb7e0357089277d/src/slash-bedrock/libexec/brl-fetch#L405-L451).
> 
> The issue here wasn't your original `usermod` within the chroot - that was fine. The issue was `import_users_and_groups()` only added users and groups, but didn't add users _to_ groups.
> 
> What I'd like to do is revert back to your original `chroot "${target_dir}" usermod ...` strategy and expand `import_users_and_groups()` to add users _to_ groups.
> 
> Something like (untested, off the top of my head):
> 
> ```
> while read -r line; do
> 	if [ -z "$(echo "${line}" | awk -F: '{print$4}') ]; then
> 		continue
> 	fi
> 	group="$(echo "${line}" | awk -F: '{print$1}')"
> 	for user in $(echo "${line}" | cut -d: -f4- | sed 's/,/ /g'); do
> 		addgroup "${user}" "${group}"
> 	done
> done <"${group_file}"
> ```
> 
> I can go after that if you'd like, or you're welcome to take a crack at it if you prefer; no strong preference here.

Thanks for the excellent explanation! Yes, that makes a lot of sense to me. And if that was really off the top of your head, colour me impressed! :-)

cf. PR #144 which obsoletes this one.

PR #144, opened 2019-09-08
Add users to groups

Philantrop:
This addresses https://github.com/bedrocklinux/bedrocklinux-userland/pull/143#issuecomment-529057095 and obsoletes PR #143.


paradigm:
Looks good!

PR #145, opened 2019-10-09
Add Artix Linux fetch support

runningnak3d:


src/slash-bedrock/share/brl-fetch/distros/artix
```diff
@@ -0,0 +1,126 @@
...
+#
+# Copyright (c) 2016-2019 Daniel Thau <danthau@bedrocklinux.org>
+#
+
+# shellcheck source=src/slash-bedrock/libexec/brl-fetch
+. /bedrock/share/common-code
+trap 'fetch_abort "Unexpected error occurred."' EXIT
+
+check_supported() {
+	true
```
paradigm: Functionally, this controls whether the given distro is listed by `brl fetch -L` or `brl fetch -X`.  This isn't at all clear from the code, but the intent here is to indicate whether or not there's a Bedrock Linux maintainer for the given distro.  That is, [the third column here](https://bedrocklinux.org/0.7/distro-support.html) is populated.  If there isn't, we want it under `-X` so users know to expect the possibility that stuff from that distro may not work on Bedrock.  If you're interested in becoming the Artix Linux maintainer and ensuring Artix specific things work on Bedrock Linux, we can certainly discuss that possibility.  Until then, however, this should probably be `false`.

runningnak3d: Like Manjaro I will change this to false, and you can decide when / if you want to list it as supported.

paradigm:
Looks good to me, thanks!

PR #146, opened 2019-10-10
Add Manjaro Linux fetch support

runningnak3d:


src/slash-bedrock/share/brl-fetch/distros/manjaro
```diff
@@ -0,0 +1,133 @@
...
+	# mount setup.
+	sed 's/^Architecture[ \t]*=.*$//' "${1}/etc/pacman.conf" | \
+		awk -v"a=${distro_arch}" '{print} $0 == "[options]" {print "Architecture = "a}' | \
+		sed 's/^[ \t]*CheckSpace/# &/' \
+		> "${1}/etc/pacman.conf-new"
+	mv "${1}/etc/pacman.conf-new" "${1}/etc/pacman.conf"
+
+	LC_ALL=C chroot "${1}" /usr/bin/update-ca-trust
+	LC_ALL=C chroot "${1}" /usr/bin/pacman-key --init
+	LC_ALL=C chroot "${1}" /usr/bin/pacman-key --populate manjaro archlinux
```
paradigm: Trailing space.   Search for `make check` in the `Makefile` or `README.md`.

src/slash-bedrock/share/brl-fetch/distros/manjaro
```diff
@@ -0,0 +1,133 @@
...
+brl_arch_to_distro() {
+	case "${1}" in
+	"x86_64") echo "x86_64" ;;
+	*) abort "brl does not know how to translate arch \"${1}\" to ${distro} format" ;;
+	esac
+}
+
+list_architectures() {
+	
+	echo "x86_64"
```
paradigm: [Distrowatch indicates Manjaro supports aarch64](https://distrowatch.com/table.php?distribution=manjaro).  Poking around their mirrors I also see i686.  Moreover, looking under `Editions` on `manjaro.org/download/` I see `32bit` and `Arm`.  So it _seems_ to support i686 and aarch64.  However, [Wikipedia](https://en.wikipedia.org/wiki/Manjaro) indicates both i686 and ARM are unofficial.

I've not been following Manjaro at all.  Do you understand the current Manjaro ISA support situation?  I figure _either_:

- Manjaro supports more than just x86_64, and we should add `i686` and `aarch64` here.
- Manjaro Linux 32 and Manjaro Linux ARM are separate projects (like Arch Linux 32 and Arch Linux ARM) and we should add `brl fetch` support for those as well.

If you haven't already, read the "Experimental features" section of `brl fetch --help`.  You can `brl fetch` non-native ISA distros on Bedrock by leveraging `qemu`.  No need to have an ARM SBC or something to test these.

src/slash-bedrock/share/brl-fetch/distros/manjaro
```diff
@@ -0,0 +1,133 @@
...
+#
+# Copyright (c) 2016-2019 Daniel Thau <danthau@bedrocklinux.org>
+#
+
+# shellcheck source=src/slash-bedrock/libexec/brl-fetch
+. /bedrock/share/common-code
+trap 'fetch_abort "Unexpected error occurred."' EXIT
+
+check_supported() {
+	true
```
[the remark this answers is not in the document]
runningnak3d: No problem, I will update this. I had it set to ```true``` so that I could brl fetch without any additional options when I was testing.
As for supporting it, yes, I will be supporting it -- but you can decide if / when to turn it into a supported distro.

src/slash-bedrock/share/brl-fetch/distros/manjaro
runningnak3d: No problem -- thanks for the heads up on make check.

src/slash-bedrock/share/brl-fetch/distros/manjaro
runningnak3d: When I was going through their repos I saw that i686 was included, but I didn't see aarch64 (must use separate repos?). I didn't add i686 yet because I wanted to make sure x86_64 was working well first. I figured it was better to hash these things out in a PR rather than a gist hence the incomplete PR. I will be glad to go ahead and add i686 support to this PR once I verify that it is fully supported / not about to be deprecated by the Manjaro team. As for ARM support, that will have to be a seperate PR since they use separate repos.

src/slash-bedrock/share/brl-fetch/distros/manjaro
```diff
@@ -0,0 +1,133 @@
...
+	cp -a "/etc/passwd" "${target_dir}/etc/passwd"
+	cp -a "/etc/group" "${target_dir}/etc/group"
+
+	setup_chroot "${bootstrap_dir}"
+	share_cache "packages" "${bootstrap_dir}/target-root/var/cache/pacman/pkg/"
+	# -a to use the api, -P all protocols, -t 1 timeout if it takes longer than 1 second
+	LC_ALL=C chroot "${bootstrap_dir}" pacman-mirrors -a -P all -t 1
+	# Because there is no default mirrorlist in Manjaro, we have to copy it
+	mkdir -p "${target_dir}/etc/pacman.d"
+	cp "/etc/pacman.d/mirrorlist" "${target_dir}/etc/pacman.d"
```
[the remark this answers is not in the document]
runningnak3d: Yes, I meant to copy the mirrorlist from the output of pacman-mirrors. This is a copy / paste fail from my dev script into the brl fetch script. I didn't get an error on the final test before making this PR because I was in an Arch based strata.

As for not using ```pacman-mirrors``` -- no, there probably isn't a good reason to use it. I was trying to stick as close to possible to the actual install routine, and generating a mirrorlist with pacman-mirrors is part of the install. I will go ahead and fix this so that the mirror from ```"${target_mirror}" ```is used.

src/slash-bedrock/share/brl-fetch/distros/manjaro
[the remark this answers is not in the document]
paradigm: > No problem, I will update this. I had it set to true so that I could brl fetch without any additional options when I was testing.

`true`/`false` here doesn't change options required when testing.  It just changes if it shows up when one runs `brl fetch --list` versus when one runs `brl fetch --experimental`.

> As for supporting it, yes, I will be supporting it -- but you can decide if / when to turn it into a supported distro.

This doesn't parse for me.  There shouldn't be a disconnect between whether someone is officially supporting Manjaro usage in Bedrock and whether or not Manjaro is officially documented as supported in Bedrock.  It's not something I decide arbitrarily.  Maybe I'm confused by what you mean by "supporting" and what you mean by "it."

The concern with `check_supported()` is whether or not an adequately capable person is actively ensuring all notable distro-specific bits work as part of a Bedrock Linux system.  If there is, we document it [here](https://bedrocklinux.org/0.7/distro-support.html), make `check_supported()` `true`, and direct any user reported issues to the corresponding maintainer.  Users should be able to rely on this to a reasonable degree.  If not, we mark it N/A [here](https://bedrocklinux.org/0.7/distro-support.html), make `check_supported()` `false`, and apologies to users who have issues and explain that we don't have anyone with the expertise and bandwidth to remedy whatever problems they're having.

If:

- You know Manjaro very well (e.g. I've been using Debian for over a decade, so I can support Debian.  Philantrop is on the Exherbo team, so he can support Exherbo).
- You regularly, actively use Manjaro as part of a Bedrock Linux system on a day-to-day basis such that you're likely to personally run into Bedrock/Manjaro issues once they arise.
- You are reasonable active in some Bedrock communication platform (IRC, reddit, github, LQ forums) such that people who need assistance can get your attention in short order.
- You are willing and eager to undertake the associated responsibility.

we can add you to [this table](https://bedrocklinux.org/0.7/distro-support.html) and make `check_supported()` here `true`.  Otherwise, that column is `N/A` and `check_suppored()` here should remain `false` until such a person shows up.

src/slash-bedrock/share/brl-fetch/distros/manjaro
paradigm: > When I was going through their repos I saw that i686 was included, but I didn't see aarch64 (must use separate repos?).

:man_shrugging:

> I didn't add i686 yet because I wanted to make sure x86_64 was working well first. I figured it was better to hash these things out in a PR rather than a gist hence the incomplete PR.

Yep, makes sense.

> I will be glad to go ahead and add i686 support to this PR once I verify that it is fully supported / not about to be deprecated by the Manjaro team.

Sounds good.

If Manjaro doesn't support it (anymore?) and it's just vestigial or experimental in the mirrors, also check for the possibility that another, separate group is maintaining it.  This is what happened with Arch Linux: there's an Arch Linux (which is x86_64 only) and a separate group/project called Arch Linux 32 (which supports various x86 flavors).  Since they're separate groups, `brl fetch` treats them as separate distros; we have a `brl fetch arch` and a `brl fetch arch-32`.  Same thing for ARM.

> As for ARM support, that will have to be a seperate PR since they use separate repos.

I don't follow why that requires them to be separate PRs.  It should be quite possible to make them one PR:

Ubuntu uses different repos for x86_64 and aarch64, but since they're all Ubuntu [brl fetch implements them in the same file](https://github.com/bedrocklinux/bedrocklinux-userland/blob/fc89de933584a188a33c41f813ba49ca31de7d79/src/slash-bedrock/share/brl-fetch/distros/ubuntu#L25-L37).  One could `git add -p` and break them up into different PRs, but that seems like an unnecessary pain.

[x86_64 Arch Linux](https://github.com/bedrocklinux/bedrocklinux-userland/blob/fc89de933584a188a33c41f813ba49ca31de7d79/src/slash-bedrock/share/brl-fetch/distros/arch) and (aarch64 Arch Linux)[https://github.com/bedrocklinux/bedrocklinux-userland/blob/fc89de933584a188a33c41f813ba49ca31de7d79/src/slash-bedrock/share/brl-fetch/distros/arch-arm] are maintained by different projects/teams (they're _technically_ different distros), so they're separate files.  Where I to write them from scratch today, I'd make them one PR with two different files in it.

It'd be easier for me to reasonable about them if they're one PR, as it'd avoid having to communicate about what are similar issues across different PRs.  However, if for whatever reason that's easier for you, I could probably manage it.

runningnak3d: No, one PR works for me :)

With that said, i686 is supported by a separate team, but they use the official repos. So -- with that in mind, do you want a seperate file for i686 support, or since they use the official repos, include it here? @paradigm

paradigm: I'm not sure I follow the distinction you're making.  If the Manjaro folks are doing all the work to maintain the i686 repository, how is another team involved?  What is this other team doing?

The concern here is around who the user trusts.  If a user runs `brl fetch <distro>`, we can assume they trust that distro's developers.  If someone runs `brl fetch -a i686 manjaro` they should expect to get packages from the Manjaro developers.  If a user runs `brl fetch -a i686 manjaro-32`, he or she should be aware that they're getting something from a separate Manjaro 32 folks and trust that party.  However, if it's the same group of people responsible for both x86_64 and i686 flavors of the same distro, it should be the same `brl fetch` distro with the `-a` flag differentiating the architecture.

If you're very familiar with Manjaro yourself and just didn't get the message across to me, we can keep discussing it.

If the issue here is that _neither_ of us knows with high confidence exactly what's going on with the non-x86_64 Manjaro flavors, I'd rather just refrain from adding the non-x86_64 Manjaro items to `brl fetch`.  Given limited support/maintenance manpower, completion-ism for its own sake is actively detrimental.  Most of my Bedrock development time as of late has been chasing distro updates that break `brl fetch` [1](https://old.reddit.com/r/bedrocklinux/comments/divzmu/error_when_fetching_ubuntu/) [2](https://old.reddit.com/r/bedrocklinux/comments/dksot8/trying_to_fetch_centos_8_but_brl_fetch_uses_the/) [3](https://old.reddit.com/r/bedrocklinux/comments/dowfc0/fedora_31_fetch_issue_workaround/).  It may be more worthwhile to wait until someone familiar with the project explicitly requests the feature.

runningnak3d: So, there is a separate dev team that is working on Manjaro32: https://manjaro32.org but they have access to the official Manjaro repos, and they use the official Manjaro forums, but that is where the overlap ends.

Since you base whether it is a fetchable arch for a distro, or a seperate fetch (arch-32, vs arch -a i686) on the support that a user can expect to get, this will need to be a seperate fetch since they are using Arch32 as their upstream. 

With that said, adding the support is trivial, so you can decide if you want to include them. If not, no biggie, no real time was wasted.

paradigm: Instead of continuing to fret over specifics here, I'll just pull this as-is for now and we can consider adding the other instruction set architectures if and when there's a need.

paradigm:
Many thanks for your work to add support for Manjaro!

PR #150, opened 2019-12-01
Fix typo in brl-disable

person4268:
Because of an incorrect variable name, brl-disable has this output if a stratum doesn't exist:
```
/bedrock/libexec/brl-disable: line 45: stratum: parameter not set
```


paradigm:
Nice catch, thanks!

PR #156, opened 2020-01-30
pmm: Implement list upgradable packages

NICHOLAS85:
This is going to be updated as I have the time to test and implement it across all the package managers. Currently, it simply prints the package name prepended by <stratum>:<package-manager>. ie:
```sh
$ # Apt user interface
$ pmm list --upgradable
* strat -r debian apt-get -smV --no-download dist-upgrade | awk '/=>/{print$1}'
debian:apt      qt5-default
debian:apt      qt5-qmake
debian:apt      qt5-qmake-bin
debian:apt      qtbase5-dev
debian:apt      qtbase5-dev-tools
* strat -r arch pacman -Qu | cut -d' ' -f1
arch:pacman     python
arch:pacman     python-setuptools
arch:pacman     python-six
arch:pacman     python2-setuptools
arch:pacman     python2-six
```
A question I have is whether or not version information should be included as well and how that should be standardized. (`0.84.8 => 0.90` or `0.84.8 -> 0.90`, etc)


paradigm:
Looks good so far!

> A question I have is whether or not version information should be included as well and how that should be standardized. (0.84.8 => 0.90 or 0.84.8 -> 0.90, etc)

I did some testing in pre-beta `pmm` outputting version information in a third column for various commands to better match expectations from commands like `dpkg -l` and `pacman -Q`.  I found:

- It added more complexity than I'd like.
- It made the corresponding implementations too long to reasonably display in the ui.  Most ended up looking like the `print-package-version` or `cache-package-db` items commonly do.
- In some cases it added runtime overhead, as I had to use much slower implementation commands to gather version information for every package.
- In my own personal workflow I very rarely actually use version columns in such commands and it didn't actually give me much.  If I care about the version I query package information directly.  I recognize others may operate differently here.
- It opens a can of worms for what else to output.  Do we also output instruction set architecture the way `dpkg -l` does?  Package descriptions?  `USE` flags?

I concluded it's not worthwhile, and dropped it from every operation.  I'm inclined to just continue the pattern of leaving it out of this one as well.  That having been said, I'm open to counter arguments if you disagree.

NICHOLAS85:
We'd have to wait and see if people's expectations warrant a change in behavior, but for now, from your points, I'd agree that it can be left out.

NICHOLAS85:
Currently waiting on some updates to hit my machine to test implementing this for the xbps package manager. This was a fun little project, I learned a lot about how these other package managers work! I hope you find the current implementations adequate. I'm going to be looking into the package name filtering to make sure they're foolproof.

paradigm:
I like everything I'm seeing so far.  It's nice to see that `pmm`'s configuration design is intuitive enough that others can pick it up without documentation or hand holding.  I'm glad you're enjoying working on it.  There's no rush; if this isn't ready by the time `pmm` leaves beta we can always just make it a `0.7.x` update following.  Nothing is blocked on it.

NICHOLAS85:
Implemented in all package managers now! The only issue I could see is in portage and xbps when printing packages which contain version info in their names (Due to using `-[0-9]` as a delimiter). For example with xbps `boost-python3-1.72-1.72.0_1` prints as `boost-python3` instead of `boost-python3-1.72`. I have only been able to find this example in Void repos and none so far in Gentoos so it may be an insignificant issue. If you know of a better way of filtering for those feel free to share.

NICHOLAS85:
I've used a different filter which cuts everything after the last `-[0-9]` match in xbps and prints between the first `/` and last `-[0-9]` in portage. That should cover the issue I pointed out ealier.

NICHOLAS85:
Noting the following commands have issues when placed into the pmm config files
```
sed 's/.*\/\(.*\)-[0-9].*/\1/;t;d' 
sed 's|\(.*\)-.*|\1|'
```
Their current `sed -E` adaptations can't be used.

NICHOLAS85:
Closing for now, will come back once `--standardize/--pass-through` is being looked at again as a pmm flag.

PR #176, opened 2020-05-24
_brl: Fix brl-strat command completion

NICHOLAS85:
`brl strat` had an issue where it wouldn't complete after the third argument. This allows it to complete correctly after that.


paradigm:
Ah, that makes sense.  Nicely done, thanks!

Titaniumtown:
Oh thanks! I thought this issue was just me. lol

PR #182, opened 2020-05-27
Manjaro-arm support

Titaniumtown:
Things I've done in this PR: implement manjaro-arm support

Manjaro-arm support: I've added manjaro-arm support. But I've hit a roadblock, there is no manjaro-tools-base or manjaro-arm-tools package on manjaro arm. When you run `brl fetch manjaro` that installs and uses the manjaro-tools-base package for installing packages into the new root directory.
Worst case scenerio, if I can't figure out how to do it, I can just install arch linux arm version of those packages (see https://wiki.manjaro.org/index.php?title=Manjaro-tools#Files) and do it that way.
Anyone know how I can proceed?


Titaniumtown:
Bump, anyone know how I should proceed?

Titaniumtown:
Bump. Any tips?

paradigm:
Is this (a) an official ARM release of the "normal" Manjaro project or (b) an unofficial project not run by the Manjaro devs?  If it's (a), we'll want to make it an extension of the existing Manjaro fetch code.  If it's (b), you're correct here in making it an independent project.

> But I've hit a roadblock, there is no manjaro-tools-base or manjaro-arm-tools package on manjaro arm. When you run `brl fetch manjaro` that installs and uses the manjaro-tools-base package for installing packages into the new root directory.

[Note there are no Bedrock Manjaro maintainers](https://bedrocklinux.org/0.7/distro-compatibility.html).  If you want to contribute something like this - and I absolutely encourage you to do so - you'll have to figure this part out yourself.  Copy/pasting existing `brl fetch` code is easy; the interesting part is figuring out how to adapt it for the distro you're interested in.

The existing `brl fetch manjaro` code uses the [double bootstrap strategy](https://bedrocklinux.org/0.7/extending.html#brl-fetch-new-distros-double-bootstrap).  We know Manjaro provides a `basestrap` binary which bootstraps a Manjaro install.  Given this, what `brl fetch manjaro` does is first bootstrap `basestrap`, then have `basestrap` bootstrap the actual distro.

If Manjaro ARM provides `basestrap`, you'll want to figure out which package provides it.  It could be a different package name from what Manjaro proper uses.  You might also have to find unlisted dependencies; that happens sometimes.

If Manjaro ARM doesn't provide `basestrap` but does provide another tool which does the same general idea with different specifics (e.g. `pacstrap`), you'll want to its corresponding package.  You might also have to find unlisted dependencies; that happens sometimes.  You'll also need to adjust the later `basestrap` call accordingly.

If Manjaro ARM does not provide any such bootstrap tool, you may need to use [another fetch strategy](https://bedrocklinux.org/0.7/extending.html#brl-fetch-new-distros-strategies).  It's not impossible no currently-in-use `brl fetch` works for Manjaro ARM and you'll have to develop another yourself.

> Worst case scenerio, if I can't figure out how to do it, I can just install arch linux arm version of those packages (see https://wiki.manjaro.org/index.php?title=Manjaro-tools#Files) and do it that way.

I'm not sure if this fact is properly documented anywhere, but we don't allow this.  Some concerns include:

- I'm not sure if one distro's mirror would be upset we're using its bandwidth for another distro's purposes, and I don't want to risk it.  Using Arch's mirror for Arch things seems reasonable, but using Arch's mirror for Manjaro things might make someone upset.
- Bedrock tends to attract strongly opinionated users, and such a user might not want to run software from a given distro.  If a user runs `brl fetch manjaro-arm`, he or she presumably trusts the Bedrock developers and Manjaro ARM developers but may hypothetically dislike or distrust the ALARM devs and not want their code running on their machine.

Titaniumtown:
It's made by the devs, but it's a seperate distro I believe. I will look into your other suggestions. thx for the feedback

Titaniumtown:
Btw, I don't really use manjaro myself, but I just wanted to add this. (btw I use arch)

parkerlreed:
arch-install-scripts is still available on Manjaro ARM https://mirror.clarkson.edu/manjaro/arm-stable/extra/aarch64/arch-install-scripts-23-2-any.pkg.tar.xz

What's the point in their replacement package?

Titaniumtown:
Oh, I didn't see that. Thanks!

PR #184, opened 2020-06-12
Create fedora-rawhide

Fermats-Last-Account:
Added brl fetch support for Fedora Rawhide. I took regular Fedora's bootstrap code and made a few changes in order to have it work for Fedora Rawhide.

### DIFF
--- a/src/slash-bedrock/share/brl-fetch/distros/fedora-rawhide
+++ b/src/slash-bedrock/share/brl-fetch/distros/fedora-rawhide
@@ -1,0 +1,153 @@
+#!/bedrock/libexec/busybox sh
+#
+# Fedora Rawhide bootstrap support
+#
+#      This program is free software; you can redistribute it and/or
+#      modify it under the terms of the GNU General Public License
+#      version 2 as published by the Free Software Foundation.
+#
+# Copyright (c) 2016-2020 Daniel Thau <danthau@bedrocklinux.org>q
+#
+
+# shellcheck source=src/slash-bedrock/libexec/brl-fetch
+. /bedrock/share/common-code
+trap 'fetch_abort "Unexpected error occurred."' EXIT
+
+check_supported() {
+	true
+}
+
+speed_test_url() {
+	echo "development/rawhide/Everything/${distro_arch}/os/repodata/repomd.xml"
+}
+
+list_mirrors() {
+	wget -O- "https://mirrors.fedoraproject.org/mirrorlist?repo=fedora-rawhide&arch=${distro_arch}" 2>/dev/null |
+		sed -e 's,/$,,' |
+		sed 's,/[^/]*/[^/]*/[^/]*/[^/]*/[^/]*$,,'
+}
+
+brl_arch_to_distro() {
+	case "${1}" in
+	"aarch64") echo "aarch64" ;;
+	"armv7hl") echo "armhfp" ;;
+	"ppc64le") echo "ppc64le" ;;
+	"s390x") echo "s390x" ;;
+		# Mirrors call it i386, but repo calls it i686.
+	"i686") echo "i386" ;;
+	"x86_64") echo "x86_64" ;;
+	*) abort "brl does not know how to translate arch \"${1}\" to ${distro} format" ;;
+	esac
+}
+
+list_architectures() {
+	cat <<EOF
+aarch64
+armv7hl
+ppc64le
+s390x
+i686
+x86_64
+EOF
+}
+
+default_release() {
+	echo "rolling"
+}
+
+list_releases() {
+	echo "rolling"
+}
+
+determine_package_manager() {
+	if [ -e "${bootstrap_dir}/brldb/depends/dn" ] && grep -q '^dnf\>' "${bootstrap_dir}/brldb/depends/dn"; then
+		echo "dnf"
+	else
+		echo "yum"
+	fi
+}
+
+bootstrap_deps() {
+	echo "${package_manager} rpm fedora-release filesystem"
+}
+
+fetch() {
+	suffix="development/rawhide/Everything/${distro_arch}/os"
+
+	step "Downloading package information database"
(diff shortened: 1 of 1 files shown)


Titaniumtown:
shouldn't it just be another "release" of fedora?

Fermats-Last-Account:
Ideally that would be the case, however Rawhide has some different links than regular Fedora.

For example, it uses `linux/development/rawhide/Everything` as opposed to `linux/releases/32/Everything`.

I wasn't sure how to implement those changes without breaking fetch support for regular Fedora.

Titaniumtown:
Oh ok.

paradigm:
I agree with Titaniumtown.  This should just be a release of Fedora.

> For example, it uses `linux/development/rawhide/Everything` as opposed to `linux/releases/32/Everything`.

If this string is used once, just use an `if`/`else` check if the version is rawhide and handle both branches accordingly.

If it's read multiple times, populate a variable early on with an `if`/`else` and use the variable in the various locations.

> It also isn't listed in the directory Bedrock uses to find Fedora's releases. https://download-ib01.fedoraproject.org/pub/fedora-secondary/releases/

Ideal solution would be to find another way to figure out Fedora's releases which would properly handle the hypothetical scenario they drop `rawhide`.  However, I don't foresee them doing that any time soon.  It's probably safe enough to hardcode `rawhide` in the output, in addition to the dynamically fetched values.

Titaniumtown:
bump, is this going to to be added?

Titaniumtown:
Found one issue: when I do dnf update, I get "Failed to set locale, defaulting to C.UTF-8" this doesn't happen with a fedora 32 stratum.

Bobbbay:
@Titaniumtown Interesting, try `cat /bedrock/strata/{name-of-strat}/etc/locale.conf`

Locale problems generally seem small but have huge impacts on tons of applications.

Titaniumtown:
Lemme do that right now @Bobbbay Give me a second.

Titaniumtown:
I was refetching it bc I deleted the stratum, and I got the error:
```
warning: /target-root/var/cache/dnf/rawhide-2d95c80a1fa0a67d/packages/acl-2.2.53-8.fc33.x86_64.rpm: Header V4 RSA/SHA256 Signature, key ID 45719a39: NOKEY
Fedora - Rawhide - Developmental packages for the next  0.0  B/s |   0  B     00:00
Curl error (37): Couldn't read a file:// file for file:///etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-x86_64 [Couldn't open file /etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-$releasever-x86_64]
```

Titaniumtown:
It fails to fetch. ummmm

Bobbbay:
@Titaniumtown That's better, now it doesn't fetch at all! :))))

Are you sure that's the command that you ran? Try `history | grep "brl fetch"` to see all of your fetches. I see a `--experimental` flag in fetch too, but seems unrelated.

Titaniumtown:
well, I ran:
```
sudo brl fetch fedora-rawhide -n fedora-rawhide
```

Titaniumtown:
same error. hmmmm

Bobbbay:
@Titaniumtown Well then, let's break down the error.

Line one is a warning, let's ignore that *for now*.

Line two is regular curl output, at 0 bytes

Line 3 is the real deal. Shize, I think I just found it. Typo in the install code, says $releasesever instead of $releasese**r**ver. Give me a sec to find where the install files are stored, and I'll give you a fix.

Titaniumtown:
Ok, got it. Strange because it worked before, now it fails to fetch :thinking:

Bobbbay:
Yes, now that I think of it that could be wrong. 

Seems like this was an actual problem with Fedora at a time, see [this](https://ask.fedoraproject.org/t/i-cant-update-fedora-32-curl-error-37/8315), [this](https://groups.google.com/g/qubes-users/c/CKG5btVhb2Y) and [this](https://forums.fedoraforum.org/showthread.php?321917-Curl-Error-(37)-with-yum-update) (or just search up "curl error (37)")

I can't seem to find where the file that stores this is, both in this repo or online, @paradigm will know.

Bobbbay:
@Titaniumtown Could you try now? It's been a while, and developer's luck might let us fix that at a later point :laughing:

kskeigrshi:
This issue seems to have been inactive for an year. Also Fedora 35 is not added to brl fetch, maybe adding that will be a reasonable choice.

paradigm:
> This PR seems to have been inactive for an year.

Yes, that happens.  Bedrock is a _huge_ project, with tons to do, and no where enough people to get it all done in a timely manner.  Given the availability of `brl import`, expanding `brl fetch` isn't a particularly high priority.  If no one else implements this, I'll likely do so myself eventually, but after a very long list of higher priority items.

> Also Fedora 35 is not added to brl fetch, maybe adding that will be a reasonable choice.

`brl fetch fedora -r 35` works for me.  If it is not working for you, please provide more information, such as exactly what happens when you try it.

Apparently the mirror `brl fetch` was using to dynamically determine the current list of Fedora releases is not being maintained.  As in our previous discussion, I have a fix in master that will trickle down to the next beta and stable release.

For the time being, open up `/bedrock/share/brl-fetch/distros/fedora` with root permissions and either manually apply the changes shown [here](https://github.com/bedrocklinux/bedrocklinux-userland/commit/491861fb8349dc5c97bcb61ffdd782131742e89f) or just outright overwrite the file with the [this](https://github.com/bedrocklinux/bedrocklinux-userland/blob/491861fb8349dc5c97bcb61ffdd782131742e89f/src/slash-bedrock/share/brl-fetch/distros/fedora) (make sure to click "Raw").

kskeigrshi:
I've succeded in installing rawhide with `sudo dnf upgrade --refresh;sudo dnf install dnf-plugin-system-upgrade;sudo dnf system-upgrade download --releasever=rawhide;sudo dnf system-upgrade reboot` in the bedrock stratum and booting the PC with fedora's init system.

PR #186, opened 2020-06-23
InstantOS support

Titaniumtown:
I am basing this off of the arch linux install script. I still have some issues to fix, but expect support soon!

### DIFF
--- a/src/slash-bedrock/share/brl-fetch/distros/instantOS
+++ b/src/slash-bedrock/share/brl-fetch/distros/instantOS
@@ -1,0 +1,126 @@
+#!/bedrock/libexec/busybox sh
+. /bedrock/share/common-code
+trap 'fetch_abort "Unexpected error occurred."' EXIT
+
+check_supported() {
+	true
+}
+
+speed_test_url() {
+	echo "core/os/${distro_arch}/core.db.tar.gz"
+}
+
+list_mirrors() {
+	# All mirrors:
+	# mirror_list_url='https://www.archlinux.org/mirrorlist/all/'
+	# Server-side filtered list of mirrors:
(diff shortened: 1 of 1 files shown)


Titaniumtown:
I'm getting an issue, i don't know what's causing it: 

`[10/19 ( 52%)] Downloading bootstrap packages
[\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\] 100%
If you let brl-fetch automatically choose a mirror, it may have chosen a bad one.  Try manually providing a mirror with --mirror`

Titaniumtown:
It seems like this distro is actually manjaro based, so I'll change this fetch script to reflect that.

Titaniumtown:
Figured out the issue, something to do with checksums and the like.

paradigm:
While I don't think it's documented anywhere, we don't currently allow using one distro's mirrors to fetch another distro.  The main concern here is the possibility that one distro's mirror would be upset we're using its bandwidth for another distro's purposes, and I don't want to risk it.  Using Arch's mirror for Arch things seems reasonable, but using Arch's mirror for InstantOS things might make someone upset.

I know this rules out distros whose files are supersets of other distros, like Mint is to Debian.  We don't have a firm policy to handle this at this point in time.  We might loosen this `brl fetch` restriction later, or we might make a different tool to automate converting a "base" distro into a superset one; not sure yet.

Historically there hasn't been much interest amongst Bedrock folks to combine parts of multiple superset distros.  Given this, most people just hijack the one superset distro they're interested in and `brl fetch` whatever else they want.  There isn't much pressing need for `brl fetch` support of superset distros.

Titaniumtown:
ok, so should this pr just be closed then?

paradigm:
I don't see a preferable alternative.

PR #195, opened 2020-08-26
common-code: Ensure color_priority is always set

cptpcrd:
Previously, it was not set if `brl_color` was false. This breaks the installer when either a) its stdout is not a tty or b) `miscellaneous.color` is not set to true.


paradigm:
Nice catch, lgtm

PR #198, opened 2020-08-31
Add pip pmm support (draft)

the-cobalt:
First draft of attempting to add pip support for pmm. Any ways I could improve it?

### DIFF
--- a/src/slash-bedrock/share/pmm/package_managers/pip
+++ b/src/slash-bedrock/share/pmm/package_managers/pip
@@ -1,0 +1,131 @@
+#!/bedrock/libexec/busybox awk -f
+#
+# Package Manager Manager pip support script
+#
+#      This program is free software; you can redistribute it and/or
+#      modify it under the terms of the GNU General Public License
+#      version 2 as published by the Free Software Foundation.
+#
+# Copyright (c) 2013-2020 Daniel Thau <danthau@bedrocklinux.org>
+
+system_package_managers["pip"]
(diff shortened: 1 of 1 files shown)


paradigm:
Seems like a good start.

For things like

```
implementations["pip", "remove-packages-limited"]  = "strat -r ${stratum} pip ${flags} remove ${items}"
implementations["pip", "remove-packages-full"]     = ""
```

[you'll want to populate both](https://bedrocklinux.org/0.7/extending.html#pmm-new-package-managers-implementations).  Consider what happens if a user runs something like `pmm purge <pip-package>` here in contrast to what the user probably wants.

For

```
implementations["pip", "search-for-package-by-name"]   = "" # Can't figure out how to
implementations["pip", "search-for-package-by-all"]    = "" # differentiate between the two
```

I think you can

```
strat -r ${stratum} pip search ${items} | awk '{print$1}'
```

for the first and

```
strat -r ${stratum} pip search ${items} | awk '{print$1}' | grep ${items}
```

for the latter.  If you figure out how to list available packages, `| grep ${items}` on it might be better or worse depending on how pip's search and list system works.

I don't know pip's nuances at all for things like populating other fields or explicit vs implicit handling.

the-cobalt:
Thanks for responding!
For `"reinstall-packages"` implementation, would it be clunky to use `pip uninstall <pkg> && pip install <pkg>`?
Also, I added your suggestions.

paradigm:
I'd prefer that to leaving it blank and confusing users when they try it.

PR #205, opened 2020-10-28
Add locale support for Gentoo strat installer

### LINKED ISSUE #204: Gentoo strat: every time you use emerge it outputs a locale error
This is more of an aesthetic issue than anything, but every time emerge is used this gets printed:
`setlocale: unsupported locale setting`
A fix for this is to tamper with your locale, but having it work out of the box would be better.

Bobbbay:
The Gentoo installer does not support locales very well - it seems to not be setting `/etc/locale.gen` very well. This PR adds support for a sexy `locale.gen` file, supporting `en_US ISO-8859-1`, `en_US.UTF-8 UTF-8`, and `C.UTF-8 UTF-8`. 

Fixes #204

### DIFF
--- a/src/slash-bedrock/share/brl-fetch/distros/gentoo
+++ b/src/slash-bedrock/share/brl-fetch/distros/gentoo
@@ -111,7 +111,6 @@ fetch() {
 	cat>>${target_dir}/etc/locale.gen<<EOF
 	en_US ISO-8859-1
 	en_US.UTF-8 UTF-8
-	C.UTF-8 UTF-8
 	EOF
 	chroot "${target_dir}" locale-gen
 	chroot "${target_dir}" eselect locale set 5
--- a/src/slash-bedrock/share/brl-fetch/distros/gentoo
+++ b/src/slash-bedrock/share/brl-fetch/distros/gentoo
@@ -107,4 +107,12 @@ fetch() {
 	mkdir -p "${target_dir}/etc/portage/repos.conf"
 	cp "${target_dir}/usr/share/portage/config/repos.conf" "${target_dir}/etc/portage/repos.conf/gentoo.conf"
 	LC_ALL=C chroot "${target_dir}" emerge-webrsync
+	echo "Defaulting to en_US ISO-8859-1 and en_US-UTF-8 UTF-8"
+	cat>>${target_dir}/etc/locale.gen<<EOF
+	en_US ISO-8859-1
+	en_US.UTF-8 UTF-8
+	C.UTF-8 UTF-8
+	EOF
+	chroot "${target_dir}" locale-gen
+	chroot "${target_dir}" eselect locale set 5
 }


Bobbbay:
Although I do have some problems with this PR - mainly that it sort of forces on some locales that the user might not want. On the other hand, I don't want to ask the user what locales they want during the install, because that would be a bumpy ride for many. 

(for reference, these are the default locales):
```
en_US ISO-8859-1
en_US.UTF-8 UTF-8
C.UTF-8 UTF-8
```

paradigm:
I applaud the initiative here.  However, I'm in agreement with:

> Although I do have some problems with this PR - mainly that it sort of forces on some locales that the user might not want.

Bedrock really tries to minimize forcing these kinds of things on users.  We want to thoroughly explore the situation and ensure there's no alternative before resorting to this kind of thing.

> On the other hand, I don't want to ask the user what locales they want during the install, because that would be a bumpy ride for many.

If Bedrock is installed via hijacking another install, [it detects and configures the system to use the hijacked install's localegen values.](https://bedrocklinux.org/0.7/configuration.html#locale-localgen)  No need for Bedrock to ask the user for these things; we just let whatever installer the user explicitly chose to use do it for us.

-----

If you were closer to an agreeable fix I'd have worked with you to upstream your fix, but as it is the actual fix here was pretty distant.  I've implemented it in 0.7.19beta1, which removes the need for this PR, and so I'm closing it.  However, I don't want to be discouraging; do feel free to try similar PRs in future situations.  If you're itching to help but don't know how, let me know and I can work with you to find something suitable.

Bobbbay:
Hey @paradigm! 

Thanks for the applause. 

Just read up on #204. Too bad it's not something this PR fixed 😜. 

> Bedrock really tries to minimize forcing these kinds of things on users. We want to thoroughly explore the situation and ensure there's no alternative before resorting to this kind of thing.

Yes, I really agree with that philosophy. Thought as much.

> If Bedrock is installed via hijacking another install, it detects and configures the system to use the hijacked install's localegen values. No need for Bedrock to ask the user for these things; we just let whatever installer the user explicitly chose to use do it for us.

Ah, that is sexy to knows, cheers!

I'd most definitely be interested in further development in Bedrock Linux. *Please* do let me know what I can do!

Cheers, Bobbbay

paradigm:
Think you could help with these research efforts?  Maybe the Gentoo/OpenRC bits to start, then if you're keen on it you could dig into other distros. https://github.com/bedrocklinux/bedrocklinux-userland/issues/207 https://github.com/bedrocklinux/bedrocklinux-userland/issues/208

Bobbbay:
One step ahead of you, already commented on 207 :wink:.

I'll be commenting on both of those in the future. Time for some research on 208. Thank you, and see ya' around!

the-cobalt:
@Bobbbay 

These lines seem odd to me:
```
+	chroot "${target_dir}" locale-gen
+	chroot "${target_dir}" eselect locale set 5
 }
```
Not 100% sure how the eselect code works, but it feels like hard-coding the locales location in the list might not be the best idea. Maybe that's not how it works though! Maybe it is in the same place everytime! Just stroke me as odd.

Bobbbay:
@the-cobalt well, if you mean using a different file for locales, I assume it's possible (if not, just patch it, praise Portage for working beautifully with patches). Then again yet another reason this issue is closed :)

PR #209, opened 2020-11-11
pmm: dnf: Fix assume-yes flag

NICHOLAS85:
dnf uses `--assumeyes` instead of `--yes` for their long flag


NICHOLAS85:
Merged fix with #215

PR #212, opened 2020-11-13
Minor feature: Add -a/--all option to `brl which` to list all strata that provide a given binary

cptpcrd:
## Rationale

Sometimes, when on a Bedrock system, I find myself wondering "which strata do I have X installed in?" What I'd like is a combination of the functionality of `brl which -b` (which prints the stratum that provides a given binary) and `which -a` (with GNU coreutils, this prints all the locations in which the given program can be found). I could probably implement this with a quick one-liner using `strat -r` and `which`, but to me this feels like a missing feature.

In a nutshell, here's what I'm aiming for:
```
$ brl which -a xbps-install
void
void-musl
```

This PR adds a `-a`/`--all` option to `brl which` which does this.

## Notes

1. This is mostly a proof-of-concept. I'm not sure if it will work in all scenarios, and I'm happy to completely rewrite it if necessary to fix bugs or make it follow better practice. I'm not in any rush to get this merged; this is just something I'd kind of like to see eventually that I think other people would find useful too.
2. There's an argument to be made that this should also work like `which_file` if any of the arguments contain a `/`. I can implement that too.
3. The interface might need changes; right now if multiple arguments are passed it's unclear which strata names correspond to which arguments.

### DIFF
--- a/src/slash-bedrock/libexec/brl-which
+++ b/src/slash-bedrock/libexec/brl-which
@@ -24,6 +24,7 @@ Options:
   ${color_cmd}-f${color_norm}, ${color_cmd}--file      ${color_norm}which ${color_term}stratum${color_norm} provides a given file path
   ${color_cmd}-p${color_norm}, ${color_cmd}--pid       ${color_norm}which ${color_term}stratum${color_norm} provides a given process ID
   ${color_cmd}-x${color_norm}, ${color_cmd}--xwindow   ${color_norm}which ${color_term}stratum${color_norm} provides a given X11 window (requires xprop)
+  ${color_cmd}-a${color_norm}, ${color_cmd}--all       ${color_norm}search in all ${color_term}strata${color_norm} (files and binaries only)
   ${color_cmd}-h${color_norm}, ${color_cmd}--help      ${color_norm}print this message
 
 Guessing reasoning for ${color_sub}<blank>${color_norm} based on identifier:
@@ -67,6 +68,19 @@ which_bin() {
 	done
 }
 
+which_bin_all() (
+	min_args "${#}" "1"
+
+	while [ -n "${1:-}" ]; do
+		for stratum in $(list_strata); do
+			if path="$(strat -r "${stratum}" /bedrock/libexec/busybox which "$1" 2>/dev/null)"; then
+				which_file "/bedrock/strata/${stratum}/${path}"
+			fi
+		done
+		shift
+	done
+)
+
 which_file() {
 	while [ -n "${1:-}" ]; do
 		if ! [ -e "${1}" ]; then
@@ -89,6 +103,19 @@ which_file() {
 		shift
 	done
 }
+
+which_file_all() (
+	min_args "${#}" "1"
+
+	while [ -n "${1:-}" ]; do
+		for stratum in $(list_strata); do
+			if [ -e "/bedrock/strata/${stratum}/$1" ]; then
+				which_file "/bedrock/strata/${stratum}/$1"
+			fi
+		done
+		shift
+	done
+)
 
 which_pid() {
 	while [ -n "${1:-}" ]; do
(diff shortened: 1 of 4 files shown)


paradigm:
A number of people have shown interest in something like this over the years (e.g.  https://gist.github.com/NICHOLAS85/a01d5f64744e4924ec0a5d5fd112cbc1).  I've been hesitant in the past, but at some point I should probably just accept the demand for this outweighs the associated complexity.

In the past, another option I've considered as an alternative to this would be a new `brl` subcommand which runs the provided arguments for all strata.  You could then do something like

```
$ # list all strata that provide `ls`
$ brl all command -v ls
$ # read all /etc/os-release files
$ brl all cat /etc/os-release
```

However, in practice, I expect the vast majority of use cases to be equivalent to `brl which -a`, and in that case, it'd make more sense to group the functionality with `brl which`.

So I guess we'll go with `brl which -a`.  Given this, some changes I'd like to see from your proof of concept before upstreaming something like this:

- Look through `/bedrock/share/common-code` for helper functions like `list_strata` rather than writing your own.
- It should probably support both functionality for both binaries in the `$PATH` and files, since there's obvious ways to interpret it for both.
- The `-a`/`--all` flag could be something that can be combine with other flags.
	 - If you run `brl which -ab <binary>` it'll find strata that provide the given binary
	 - If you run `brl which -af file` it'll find all strata that provide the given file path if `/bedrock/strata/<stratum>` is prefixed
	 - If you _just_ run `brl which -a <arg>` it should auto-detect binary vs file vs something it doesn't know how to all-ify and error out
	 - If we ever think of a way to do this with another identifier type where all is applicable, we can use it there then
- You shouldn't have to search the `$PATH` yourself.  Other strategies to consider:
	- Loop over the strata and `strat "${strat}" /bedrock/libexec/busybox sh -c "command -v \"$cmd\""`
	- Loop over the strata and `strat "${strat}" /bedrock/libexec/busybox which "$cmd"`
- bash, zsh, and fish completion (see `/bedrock/share/{bash,zsh,fish}`)

cptpcrd:
@paradigm I've pushed changes that should address most of your concerns, though I have a few of my own. I'll leave comments at relevant points.

src/slash-bedrock/libexec/brl-which
```diff
@@ -67,6 +68,19 @@ which_bin() {
 	done
 }
 
+which_bin_all() (
+	min_args "${#}" "1"
```
cptpcrd, line 72: I didn't test this before pushing, and it appears `min_args` doesn't work reliably from inside functions. Currently, this happens, and it's not immediately obvious to me how to fix it:

```
$ brl which -ab
ERROR: Insufficient arguments, see `--help`.
````

File: src/slash-bedrock/libexec/brl-which [hunk 2]
```diff
@@ -67,6 +68,19 @@ which_bin() {
...
+
+	while [ -n "${1:-}" ]; do
```
cptpcrd, line 74: The normal `brl which` interface doesn't really work properly with `-a`. For example, this is very ambiguous:

```
$ brl which -a cmake git
void
void-musl
arch
```

Some of the formats that leapt to mind include:

```
$ brl which -a cmake git
void: cmake
void-musl: cmake
arch: git
$ brl which -a cmake git
cmake: void
cmake: void-musl
git: arch
$ brl which -a cmake git
cmake: void void-musl
git: arch
```

src/slash-bedrock/libexec/brl-which
```diff
@@ -67,6 +68,19 @@ which_bin() {
...
+		for stratum in $(list_strata); do
+			if path="$(strat -r "${stratum}" /bedrock/libexec/busybox which "$1" 2>/dev/null)"; then
```
cptpcrd, line 76: I can add handling for "file/binary not found" both here and in `which_file_all`, but what should be the behavior? Something like this?

```
$ brl which -a git 0ad cmake
arch: git
WARNING: 0ad not found in any strata
void: cmake
# exit with status 1
```

jake-87: I don't think it makes much sense to phrase it as a warning, as it's really just the absence of something. Perhaps something closer to just `0ad not found in any strata`?

cptpcrd: That would probably depend on the exact format; see [above](https://github.com/bedrocklinux/bedrocklinux-userland/pull/212/files/0ac59f6294003c1f6c0342a132e81259b1cd6218#r567808165). e.g. for some it could be `0ad not found`; for others `0ad: not found` might make more sense. Colorization might help too.

src/slash-bedrock/libexec/brl-which
jake-87: Personally
 ```
$ brl which -a cmake git
void: cmake git
void-musl: cmake
arch: git```
makes the most sens to me, as this quickly shows what has what, while still having a per distro view.

PR #214, opened 2020-11-14
[PAUSED] pmm: Add auto_escalate[] logic

NICHOLAS85:
Allows a pm to set `auto_escalate[]`. When this var is set its value will be used to populate `${root_user}` for other pms. This is useful to mimic autosudoing uis behaviors like yay for other pms.

I tested every command to check whether they needed root permissions or not to function and prepended the command which did. However, this doesn't mean I didn't miss any or miss-categorize so I'll keep checking each commands behavior as I use this pr.
The Bedrock documentation would need to be updated to reflect this new variable of course.


NICHOLAS85:
Current issues after a day of usage:
Caching package database requires root and does not currently automatically escalate. I believe this was an issue already present in pmm that hadn't made itself apparent yet. yay, the only pm and pmm ui which a user would run without privileges to update package database is affected.

I believe pmm has some internal logic for the `${unprivileged_user}` var which cuts the extra space out from the command preview when it is unpopulated, I need to replicate this behavior for `${root_user}`.

NICHOLAS85:
Placed on hold until a per-user infrastructure is developed for pmm/Bedrock as pmm will have to manage non-root per-user package caches as well.

PR #217, opened 2020-12-16
pmm: pacman,yay: Fix which-packages-provide-file

NICHOLAS85:
Fixes broken implementation.

Quick question before you merge:
pacman, Line 129 and yay, line 135
`split($(NF-1), a, \"/\");\`
You used `$(NF-1)` to get the second from the last field. I changed this to $1 as sometimes "[installed]" is included as an extra field, making `$(NF-1)` unreliable. Heres an example problematic output:
```sh
$ pacman -F zsh
extra/zsh 5.8-1 [installed]
    usr/bin/zsh
community/grunt-cli 1.3.2-2
    usr/lib/node_modules/grunt-cli/completion/zsh
...
```
Before we continue with `$1`, is there a reason you chose `$(NF-1)`? Is there sometimes an extra field before the package name?


paradigm:
Nice

Issue #219, opened 2021-03-20
Incompatibilities with Firejail


panos:
Hi, I was experimenting with the aspect of security on Bedrock Linux and tried out sandboxing. Firejail tends to fail with some applications due to permission issues.

Here's an example:

```
$ firejail --x11=xephyr --net=wlan0 openbox 

*** Starting xephyr server: "Xephyr" "-ac" "-br" "-noreset" "-screen" "800x600" "-title" "firejail x11 sandbox" ":361" ***


*** Attaching to Xephyr display 361 ***

Reading profile /etc/firejail/openbox.profile
Reading profile /etc/firejail/globals.local
Reading profile /etc/firejail/disable-common.inc
Parent pid 561816, child pid 561822

Interface        MAC                IP               Mask             Status
lo                                  127.0.0.1        255.0.0.0        UP    
eth0-561816      de:ad:be:ef:00:00  [Redacted]    [Redacted]  UP    
Default gateway [Redacted]

Warning: /sbin directory link was not blacklisted
Warning: /usr/sbin directory link was not blacklisted
Warning: cleaning all supplementary groups
Warning: Cannot confine the application using AppArmor.
Maybe firejail-default AppArmor profile is not loaded into the kernel.
As root, run "aa-enforce firejail-default" to load it.
Child process initialized in 1755.03 ms
bouncer: could not execute
    /bedrock/bin/strat
due to: execv:
: Operation not permitted

Parent is shutting down, bye...
```

panos:
When invoking commands directly (e.g. `/usr/bin/chromium` instead of `chromium`), it seems to work properly.

cptpcrd:
AFAIK this is expected behavior. `strat` needs to run with elevated permissions so it can `chroot()` to the other stratum, and Firejail (correctly) blocks this for security reasons.

Everything should still work within the same

panos:
Is it on the maintainers of Firejail to have Firejail specifically look into the /usr/bin folder of the current strat, rather than looking at the /bedrock folder as indicated by the PATH environment variable?

cptpcrd:
> Is it on the maintainers of Firejail to have Firejail specifically look into the /usr/bin folder of the current strat, rather than looking at the /bedrock folder as indicated by the PATH environment variable?

I would probably say no. However, I also realized a few things upon reflection:
1. You could make `firejail` "just work" by adding `firejail` to the `restrict` list in `bedrock.conf`. This is

[...]

PR #220, opened 2021-03-21
Restrict Firejail binaries by default

panos:
Addresses #219.

I took the liberty of adding a few more binaries that could cause further issues and shouldn't really be accessible across stratums, but I specifically left out `firejail-ui`, as it appears to be a GUI wrapper.


paradigm:
Looks good to me!

PR #233, opened 2021-07-28
Add Carbs Linux

bashelled:
Added Carbs Linux support. Carbs is a fork of KISS, so the brl-fetch file for it is a highly modified version of the KISS one. (is still being tested)

EDIT: Done testing, working so far, although it may need to be restricted, which I have not done yet. Like I said, Carbs is a fork of KISS, and both (may) find executables on the fly.


paradigm:
Looks good to me so far.

Pan00Pernicek:
seems to work fine for me

PR #236, opened 2021-09-19
pmm: Add pmm zsh completion

NICHOLAS85:
Currently doesn't list packages when relevant but provides completions for all pms and their operations/flags.


paradigm:
Very nice :)

PR #256, opened 2022-03-08
brl-fetch: add development releases of fedora

CactiChameleon9:
This PR should fix the default release being from testing issue (sorry for the 2nd pr)

I am unsure on how to show the releases, so feedback and corrections wanted there
I also hope using `tr` is alright 
And yes, fetching the fedora development releases multiple times does add on a bit of time, however it seemed to match the coding style more than using a global variable and fedora takes a while to fetch anyway (so not loads % wise)

### DIFF
--- a/src/slash-bedrock/share/brl-fetch/distros/fedora
+++ b/src/slash-bedrock/share/brl-fetch/distros/fedora
@@ -18,7 +18,14 @@ check_supported() {
 }
 
 speed_test_url() {
-	echo "releases/${target_release}/Everything/${distro_arch}/os/repodata/repomd.xml"
+
+	if [[ $(list_development | tr -d '[:space:]') == *${target_release}* ]]; then
+		url_type="development"
+	else
+		url_type="releases"
+	fi
+	
+	echo "${url_type}/${target_release}/Everything/${distro_arch}/os/repodata/repomd.xml"
 }
 
 list_mirrors() {
@@ -53,17 +60,40 @@ EOF
 }
 
 default_release() {
-	list_releases |
+	list_release |
 		sort -n |
 		tail -n1
 }
 
 list_releases() {
+
+	echo
+	echo "Release"
+	echo "---------------------"
+
+	list_release
+
+	echo
+	echo "Development"
+	echo "---------------------"
+	
+	list_development
+}
+
+list_release() {
 	download -q 'https://dl.fedoraproject.org/pub/fedora/linux/releases/' - |
 		list_links |
 		grep '^[0-9][0-9]*/$' |
 		sed 's,/$,,' |
 		sort -n
+}
+
+list_development() {
+	download -q 'https://dl.fedoraproject.org/pub/fedora/linux/development/' - |
+	list_links |
+	sed 's,/$,,' |
+	grep -v "/" |
+	sort -n
 }
 
 determine_package_manager() {
@@ -78,8 +108,14 @@ bootstrap_deps() {
 	echo "${package_manager} rpm fedora-release filesystem fedora-gpg-keys"
 }
 
-fetch() {
-	suffix="releases/${target_release}/Everything/${distro_arch}/os"
+fetch() {	
+	if [[ $(list_development | tr -d '[:space:]') == *${target_release}* ]]; then
+		url_type="development"
+	else
+		url_type="releases"
+	fi
+	
+	suffix="${url_type}/${target_release}/Everything/${distro_arch}/os"
 
 	step "Downloading package information database"
 	url="$(find_link "${target_mirror}/${suffix}/repodata/" "primary.xml.gz")"
--- a/src/slash-bedrock/share/brl-fetch/distros/fedora
+++ b/src/slash-bedrock/share/brl-fetch/distros/fedora
@@ -68,24 +68,24 @@ default_release() {
 }
 
 list_releases() {
-	list_release
-	list_development
+	(
+		list_release
+		list_development
+	) | sort -n
 }
 
 list_release() {
 	download -q 'https://dl.fedoraproject.org/pub/fedora/linux/releases/' - |
 		list_links |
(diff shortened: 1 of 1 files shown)


src/slash-bedrock/share/brl-fetch/distros/fedora
```diff
@@ -18,7 +18,14 @@ check_supported() {
 }
 
 speed_test_url() {
-	echo "releases/${target_release}/Everything/${distro_arch}/os/repodata/repomd.xml"
+
+	if [[ $(list_development | tr -d '[:space:]') == *${target_release}* ]]; then
+		url_type="development"
+	else
+		url_type="releases"
+	fi
```
paradigm, line 26: Some of Bedrock's shell script runs in environments where bash-isms like the double brackets work, but some doesn't.  A uniform style everywhere is preferred, which means avoiding bash-isms where we can.

Something like this may work:

```
if list_development | sed -e 's/^/ /' -e 's/$/ /' | grep -qF "^${target_release}$"
```

or perhaps this:

```
if ! list_development | awk -v"x=${target_release}" '$0 == x {exit 1}'
```

I haven't tested those - they may be buggy.  More proposed sources of inspiration than firm counter-proposals.  You're welcome to play with it if you have other ideas.

src/slash-bedrock/share/brl-fetch/distros/fedora
```diff
@@ -53,19 +60,42 @@ EOF
...
-	list_releases |
+	list_release |
 		sort -n |
 		tail -n1
 }
 
 list_releases() {
+
+	echo
+	echo "Release"
```
paradigm, line 71: People script around `brl fetch -R` assuming the outputs can be sanely fed into `brl fetch -r`.  Output comments like this will break such scripting.  Please remove them.

Come Naga we can make a breaking change for this kind of thing.  Feel free to bring it up during Naga's public alpha, once we get there.  In addition to differentiating between stable vs development, we can also highlight things like LTS releases in Ubuntu or somehow indicate what `brl fetch` will default to if left unspecified.

src/slash-bedrock/share/brl-fetch/distros/fedora
```diff
@@ -78,8 +108,14 @@ bootstrap_deps() {
...
 }
 
-fetch() {
-	suffix="releases/${target_release}/Everything/${distro_arch}/os"
+fetch() {	
+	if [[ $(list_development | tr -d '[:space:]') == *${target_release}* ]]; then
+		url_type="development"
+	else
+		url_type="releases"
+	fi
```
paradigm, line 116: I see this conditional is called twice.  I'm glad you didn't make this a top level variable, as network calls would be surprising at that point.

My instinct would have been to make this a function to ensure we don't change it in one place and forget to change it in the other.  If you hadn't considered doing so, please do consider it.  If you have and disagree, that's fine; I'm not that picky about it and won't block merging over it.

paradigm (changes requested):

src/slash-bedrock/share/brl-fetch/distros/fedora
CactiChameleon9: I can do that. I wasn't sure if you would prefer lots of functions (not doing very much) fewer functions but that did more, so I went with the one I would go with (an if else statement isn't very big)

src/slash-bedrock/share/brl-fetch/distros/fedora
CactiChameleon9: I figured that would be the case... I just didn't like the output it was giving without:
```
daniel@danielLaptop ~ [1]> sudo brl fetch fedora_test -R
[sudo] password for daniel: 
[cut here]
27
28
29
30
31
32
33
34
35
rawhide
36
```
If you have any suggestions to improve that (or if you are fine with it) let me know

src/slash-bedrock/share/brl-fetch/distros/fedora
CactiChameleon9: The second one seems to work, and that is a clever solution. I like it :+1: 
Just running through a full test now before committing that

CactiChameleon9:
I just noticed a small issue with fetching rawhide (it still fetches fine):
```
[13/20 ( 65%)] Running bootstrap software
Unable to detect release version (use '--releasever' to specify release version)
Fedora $releasever openh264 (From   19 kB/s |  92 kB     00:04    
Errors during downloading metadata for repository 'fedora-cisco-openh264':
  - Status code: 404 for https://mirrors.fedoraproject.org/metalink?repo=fedora-cisco-openh264-$releasever&arch=x86_64 (IP: 209.132.190.2)
  - Status code: 404 for https://mirrors.fedoraproject.org/metalink?repo=fedora-cisco-openh264-$releasever&arch=x86_64 (IP: 38.145.60.20)
  - Status code: 404 for https://mirrors.fedoraproject.org/metalink?repo=fedora-cisco-openh264-$releasever&arch=x86_64 (IP: 18.185.136.17)
Error: Failed to download metadata for repo 'fedora-cisco-openh264': Cannot prepare internal mirrorlist: Status code: 404 for https://mirrors.fedoraproject.org/metalink?repo=fedora-cisco-openh264-$releasever&arch=x86_64 (IP: 18.185.136.17)
Ignoring repositories: fedora-cisco-openh264
Last metadata expiration check: 1:01:16 ago on Tue Mar  8 01:20:47 2022.
```
I am attempting to resolve it now

src/slash-bedrock/share/brl-fetch/distros/fedora
paradigm: I agree, `rawhide`'s placement is a bit weird.  Try piping `list_release` and `list_development` into `sort`, e.g.

```
list_releases() {
	(
		list_release
		list_development
	) | sort -n
}
```

If you do this, consider removing the `sort -n`s from `list_release` and `list_development` as they would end up being redundant.

PR #257, opened 2022-03-16
brl-fetch: fix exherbo sum checking and update arch

androw:
Exherbo now provides per file sha256 checksum file.
Also some archs have been discontinued and added.

I'm also updating speed test url as the sha1sum file might disappears as it is no more used.


paradigm:
This looks good to me.  Thank you!  It should filter down into the next beta and stable releases.

PR #259, opened 2022-03-23
brl-fetch: funtoo: add new distro

androw:
Funtoo Linux is a community-developed Linux meta-distribution and evolution of Gentoo Linux
https://www.funtoo.org/Welcome


src/slash-bedrock/share/brl-fetch/distros/funtoo
```diff
@@ -0,0 +1,85 @@
...
+#
+# Copyright (c) 2022-2022 Nicolas Lorin <androw95220@gmail.com>
+#
+
+# shellcheck source=src/slash-bedrock/libexec/brl-fetch
+. /bedrock/share/common-code
+trap 'fetch_abort "Unexpected error occurred."' EXIT
+
+check_supported() {
+        true
```
paradigm, line 17: This (poorly named) function returning true indicates there's a Bedrock maintainer for the distro, but [there is no maintainer at this point](https://bedrocklinux.org/0.7/distro-compatibility.html).  If [you want to become a Bedrock Funtoo maintainer we can certainly discuss it](https://bedrocklinux.org/contributing.html#distro-maintainer), but until that's resolved this should be `false`.

paradigm (changes requested):

src/slash-bedrock/share/brl-fetch/distros/funtoo
```diff
@@ -0,0 +1,85 @@
...
+}
+
+default_release() {
+        echo "1.4-release-std"
+}                           
+
+list_releases() {
+        cat <<EOF
+1.4-release-std
+next
```
paradigm, line 56: Is it not feasible to dynamically detect the release list?  Will we have to manually update this come 1.5 or 2.0?

src/slash-bedrock/share/brl-fetch/distros/funtoo
androw: Switch to false at the moment

src/slash-bedrock/share/brl-fetch/distros/funtoo
androw: Done for both list and default!

PR #277, opened 2022-10-23
brl-fetch: update exherbo stages links

androw:
Newest exherbo stage releases are using a new name scheme including the compiler
https://dev.exherbo.org/stages/


paradigm:
Thanks!

PR #298, opened 2023-10-22
Add Chimera support

5gura:
homepage: https://chimera-linux.org/

### DIFF
--- a/src/slash-bedrock/share/brl-fetch/distros/chimera
+++ b/src/slash-bedrock/share/brl-fetch/distros/chimera
@@ -1,0 +1,75 @@
+#!/bedrock/libexec/busybox sh
+#
+# Chimera Linux bootstrap support
+#
+#      This program is free software; you can redistribute it and/or
+#      modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


src/slash-bedrock/share/brl-fetch/distros/chimera
```diff
@@ -0,0 +1,75 @@
...
+# shellcheck source=src/slash-bedrock/libexec/brl-fetch
+. /bedrock/share/common-code
+trap 'fetch_abort "Unexpected error occurred."' EXIT
+
+check_supported() {
+	false
+}
+
+speed_test_url() {
+	echo "current/main/${distro_arch}/APKINDEX.tar.gz"
```
JamiKettunen, line 21: If @wezm doesn't mind https://www.wezm.net/v2/posts/2024/australian-chimera-linux-mirror/ could also be handled here even though it's not officially in any docs yet :)

wezm: Fine with me :-)

Issue #303, opened 2024-06-07
bwrap broken


tbodt:
```
sudo brl fetch debian -r testing -n test
sudo strat test apt install bubblewrap
strat test bwrap --dev-bind / / true
```
gives
```
bwrap: No permissions to create new namespace, likely because the kernel does not allow non-privileged user namespaces. See <https://deb.li/bubblewrap> or <file:///usr/share/doc/bubblewrap/README.Debian.gz>.
```
However, bwrap installed in the arch linux arm stratum I have (from hijacking, used as init) worked fine.

paradigm:
Bedrock has to balance making processes from different strata see different things (so they each see what they need to work without conflicting) and making them see the same thing (so they interact and everything feels like one cohesive system).  In 0.7 Poki, all processes see the same mount namespace, so they can all mount/unmount each other's items.  However, they have different root directories (`/` is local).  Ends up this upsets a sanity check constraint used by user namespaces, which is how bwrap works.  I didn't learn this until it was too late to adjust course in Poki.  For the upcoming 0.8 Naga the plan is to do per-stratum mount namespaces, which should make this and a few other situations just-work, but make other Bedrock internal implementation details more complex.

In Poki, your choices are to:

- As you found, use `bwrap` from the init-providing stratum.  This passes the user namespace sanity check constraint.
- setuid the `bwrap` executable.  This elevates its permissions and allows it to implement namespace changes the sanity check constraint, but comes at the expense of trusting `bwrap` drops permissions correctly.

I need to document this on https://bedrocklinux.org/0.7/feature-compatibility.html.

tbodt:
Thank you, I will probably use setuid. What exactly is that sanity check?

paradigm:
> Thank you, I will probably use setuid.

Happy to help

> What exactly is that sanity check?

From `man 2 unshare`:

>        EPERM (since Linux 3.9)
>               CLONE_NEWUSER  was  specified in flags and the caller is in a chroot environment (i.e., the caller's root directory does not match the root directory of the mount namespace in which it resides).

[...]

PR #304, opened 2024-06-08
strat -n

tbodt:
Behold... `strat -n bwrap` works for me fixing #303. But of course this is too much to merge into 0.7.0


paradigm:
Apologies for the delayed response.

I agree this is too much to merge into 0.7.x, and so sadly I'm closing this, but I'll definitely keep this in mind as 0.8.0 gets closer.  I strongly suspect this general approach fixes other subtle issues like GRUB not playing nicely with BTRFS/ZFS on Bedrock as a side-effect.

PR #308, opened 2024-07-27
Fixed pmm: brl not found

mertcaners12:
Fixed an error in line 103 which gives an error that says this: /bedrock/libexec/pmm: line 103: brl: not found
```
/bedrock/libexec/pmm: line 103: brl: not found
/bedrock/libexec/pmm: line 103: brl: not found
* strat -r endeavouros pacman -Sy sh: line 1: strat: command not found
ERROR: endeavouros:pacman returned 127
```

### DIFF
--- a/src/slash-bedrock/libexec/pmm
+++ b/src/slash-bedrock/libexec/pmm
@@ -97,13 +97,15 @@ strata="$(
 	done | sort
 )"
 export strata
+export PATH=$PATH:/bedrock/bin
+
 aliases="$(
-	for alias in $(list_aliases); do
-		if is_enabled "${alias}" && has_attr "/bedrock/strata/${alias}" "show_pmm"; then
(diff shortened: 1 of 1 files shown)


paradigm:
It's not immediately clear to me how one gets into the situation where `pmm` is in the `$PATH` but `brl` is missing.  That said, ensuring this works should that happen seems worthwhile.

Thinking about it, it seems possible relying on the `$PATH` here could introduce other subtle issues like this.  I'd prefer to resolve this by making the paths absolute rather than changing the `$PATH`. As far as I can see, a `/bedrock/bin/` entry only occurs once with a relative path: the `brl` call.

Your change include a presumably unintentional tabs to spaces change.

I've pulled your change to ensure you get credit here. I'll make a quick follow-up commit which both changes the indentation back and resolves the issue by making the `brl` call an absolute path.

PR #310, opened 2024-11-04
feat: chimera linux brl-fetch port

slbsh:
A `brl-fetch` target for [Chimera Linux](https://chimera-linux.org/).  
Huge thanks to the chimera people for help :)

### DIFF
--- a/src/slash-bedrock/share/brl-fetch/distros/chimera
+++ b/src/slash-bedrock/share/brl-fetch/distros/chimera
@@ -1,0 +1,77 @@
+#!/bedrock/libexec/busybox sh
+#
+# Chimera Linux bootstrap support
+#
+# Permission to use, copy, modify, and/or distribute this software for
+# any purpose with or without fee is hereby granted;
+# According to the terms of the BSD Zero Clause License.
+# 
+# Copyright (c) 2024 Anthony Slab
+#
+
+# shellcheck source=src/slash-bedrock/libexec/brl-fetch
+. /bedrock/share/common-code
+trap 'fetch_abort "Unexpected error occurred."' EXIT
+
+check_supported() {
+	false
+}
+
+speed_test_url() {
+	echo "current/main/${distro_arch}/APKINDEX.tar.gz"
+}
+
+list_mirrors() {
+   # pretty sure there's only the official one :p
+   echo "https://repo.chimera-linux.org"
+}
+
+brl_arch_to_distro() {
+	case "${1}" in
+	"aarch64") echo "aarch64" ;;
+	"riscv64") echo "riscv64" ;;
+	"ppc64")   echo "ppc64" ;;
+	"ppc64le") echo "ppc64le" ;;
+	"x86_64")  echo "x86_64" ;;
+	*) abort "brl does not know how to translate arch \"${1}\" to ${distro} format" ;;
+	esac
+}
+
+list_architectures() {
+	cat <<eof
+aarch64
+riscv64
+ppc64
+ppc64le
+x86_64
+eof
+}
+
+default_release() {
+	echo "current"
+}
+
+list_releases() {
+	echo "current"
+}
+
+fetch() {
+	step "Downloading bootstrap software"
+	url="$(find_link "${target_mirror}/apk/" "^apk-${target_arch}-.*.static")"
+   apk="$(basename ${url})"
+	download "${url}" "${bootstrap_dir}/${apk}"
+
+   download "${target_mirror}/apk/sha256sums.txt" "${bootstrap_dir}/shasums"
+   [ "$(sha256sum ${bootstrap_dir}/${apk} | tr -d' ' -f1)" = "$(grep ${apk} ${bootstrap_dir}/shasums | tr -d' ' -f1)" ] \
+      || { echo "ERROR: checksums did not match"; false; }
+
+	step "Preparing bootstrap software"
+	chmod a+rx "${bootstrap_dir}/${apk}"
+	mkdir -p "${target_dir}/etc/apk/"
+
+	step "Running bootstrap software"
+	setup_chroot "${bootstrap_dir}"
+	setup_ssl "${bootstrap_dir}"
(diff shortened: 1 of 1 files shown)


JamiKettunen:
This would supercede #298, check the recent mirror comment from me there if you want to include it here

src/slash-bedrock/share/brl-fetch/distros/chimera
```diff
@@ -0,0 +1,77 @@
...
+
+	step "Preparing bootstrap software"
+	chmod a+rx "${bootstrap_dir}/${apk}"
+	mkdir -p "${target_dir}/etc/apk/"
+
+	step "Running bootstrap software"
+	setup_chroot "${bootstrap_dir}"
+	setup_ssl "${bootstrap_dir}"
+	LC_ALL=C chroot "${bootstrap_dir}" "/${apk}" -X "${target_mirror}/${target_release}/main" -U \
+		--allow-untrusted --root "/target-root" --initdb add chimerautils base-minimal
```
JamiKettunen, line 76: Drop `chimerautils`, `base-minimal` is a smaller subset of [`base-full`](https://github.com/chimera-linux/cports/blob/master/main/base-full/template.py) which already depends on [`base-bootstrap`](https://github.com/chimera-linux/cports/blob/master/main/base-bootstrap/template.py) pulling that in implicitly

netkv:
Did quick testing, it seems to stuck at
```
[ 4/14 ( 28%)] Determining mirror
* Checking bedrock.conf [brl-fetch-mirror] items
[\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\] 100%
```
no idea why it gets stuck but
after removing the other mirror and changing list-mirrors function into
```sh
list_mirrors() {
	echo "https://repo.chimera-linux.org"
}
```
it fetches just fine and, and resulting stratum seems to work, so otherwise than that detail great job 👍🏻

slbsh:
I blame @JamiKettunen lmao, gonna undo that one

JamiKettunen:
:shrug: I can load https://au.mirror.7bit.org/chimera from EU even if latency is around a second, sounds like something the mirror tester definitely should be able to handle (no idea how it works) and subsequently skip if latency is way too high

netkv:
looking at it more, it doesn't seem to have anything with mirror (btw im in EU)
(i think if it was mirror speed issue it'd get stuck right after, with bedrock showing number of mirrors it decides from
like `* Finding fastest mirror from 2`
)
judging by `set -x` it seems like it gets stuck at `cat`
tried following and it works
```sh
list_mirrors() {
	echo "https://repo.chimera-linux.org
https://au.mirror.7bit.org/chimera
"
}
```
no idea why cat heredec doesnt work, it should achieve same result, but maybe my shell knowledge is rusty
btw does it work for you? i assume yes if that's so, that's odd, no idea why it'd be different (btw im using latest bedrock (0.7.31beta2) to test it on)

slbsh:
this is the exact way the other brl-fetch distros are done so this is really strange indeed; ill play around with it to see what I can get to work.

JamiKettunen:
now there's https://chimera-linux.org/docs/apk/mirrors btw for proper official docs :)

slbsh:
@JamiKettunen added the three from there, hopefully it works this time :p 
@netkv could you check if you can still replicate this problem with the newest changes?

netkv:
I don't understand why but it seems to work now just fine, and that's the same code, just with one mirror added

netkv:
```bash
list_mirrors() {
	cat >>EOF 
https://repo.chimera-linux.org
https://au.mirror.7bit.org/chimera
EOF
}```
```bash
list_mirrors() {
	cat <<EOF # from https://chimera-linux.org/docs/apk/mirrors
https://repo.chimera-linux.org
https://chimera.sakamoto.pl
https://au.mirror.7bit.org/chimera
EOF
}```
the redirection heredoc was reversed, that's why it was broken

netkv:
anyways works perfectly now

slbsh:
@netkv very silly, I can't believe I haven't noticed that lmao

q66:
the bootstrap process this uses is incorrect (read https://github.com/chimera-linux/chimera-live?tab=readme-ov-file#bootstrapping-the-system-with-apk, or just use chimera-bootstrap)

it also does not use package signing, the mirror list is hardcoded which sounds bad, etc

if you need any adjustments in the infrastructure to accomodate for doing these things correctly then i can do that but please don't expose users to subtly broken stuff

slbsh:
@q66 
I've already tried using the chimera-bootstrap script but bedrock does not provide getopts which the script requires. 

There are multiple problems with getting the static apk because of the version naming scheme which is not trivially sortable, and any "solution" would surely be prone to breakage.
Ideally the static apk repos could provide a `apk-something-latest.static`, or a similar way to get the latest version at a set path. 

I was not able to find any official way to get the mirrorlist, like there is on ex: arch, (other than scraping the docs page I suppose), if there is one let me know.

Someone also suggested using one of the ROOTFS images in `live/latest` which also didn't really work, although I dont exactly remember why right now.

There also is no obvious way to get the keys needed for apk reliably. I suppose they're in the cports repo, but cloning it is not possible within the brl-fetch script, nor is it possible to hard code the raw github links as preasumably the key filenames will change over time.

q66:
i've prepared some changes:

- the mirrors are now available in text format at https://repo.chimera-linux.org/mirrors.txt (each line is `URL Location (owner)`)
- the apk directory was reorganized with per-release directories and a `latest` symlink (https://repo.chimera-linux.org/apk)

as for the key names, it's a little complicated; the file names are not going to change unless something happens such as compromised builder that would result in having to drop a key; however, all of the repos except `ppc` currently use `q66@chimera-linux.org-61a1913b.rsa.pub` with the plan to switch to per-architecture keys (which are already in the repository and shipped in systems as a preparatory step) for new packages (i.e. the repos would use two keys until everything gets rebuilt); meanwhile `ppc` is a new port so it only has the per-arch key

so i need to think about it for a bit

slbsh:
I've updated the `brl-fetch` script to use the provied mirrorlist, keys, and the static apk path.

Thanks a lot for these changes and supporting such a niche use case, gotta say im def pleasantly surprised :)

PR #312, opened 2024-11-26
Add brl fetch support for OpenSUSE tumbleweed

LevitatingBusinessMan:
I tested this script quite thoroughly but I encourage others to test it as well.

It currently only has Tumbleweed support for x86_64. Adding support for Leap and other architectures should be easy, only the repository paths should have to be changed. But I currently don't have the time.

I can add pmm support later as well.

I am willing to maintain this script in the future.


netkv:
Did quick testing, can confirm it fetches properly and the resulting stratum seems to work fine
Great job 👍🏻

PR #313, opened 2024-11-29
Update bedrock.conf = Support Paths nixOS/nixpkgs

AsCuteSnow:
# Support for nix environments of both types

Hello, I added this because a user must add a manual Nix environment, but with this, there is no need to modify bedrock.conf.

I also want to add share = /nix as well, but this does not make sense if you want to do it yourself

True bedrock linux does not support nixOS but this also helps nix pmm and nixOS-xos project.


unrooted:
hi, out of curiosity, what projects you're referring to? might wanna check them out myself, quick Google search didn't show any related results

PR #318, opened 2025-03-22
Multithread stratum enabling

0x313:
This greatly improves booting speed, especially when running a system with many strata

### DIFF
--- a/src/init/init
+++ b/src/init/init
@@ -452,8 +452,10 @@ for stratum in $(list_strata); do
 		continue
 	fi
 	notice "Enabling ${color_strat}${stratum}${color_norm}"
-	/bedrock/libexec/brl-enable --skip-crossfs "${stratum}"
+	/bedrock/libexec/brl-enable --skip-crossfs "${stratum}" &
 done
+
+wait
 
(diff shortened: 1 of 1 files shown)


paradigm:
Apologies for the delayed response.

`/bedrock/libexec/brl-enable` locks almost to entire program lifetime, forcing serialization, and so my expectation is this has a negligible performance difference.  If it *does* have a performance difference, then that might indicate the locking is broken.  That said, I could be missing something.

Have you done any benchmarking and seen actual improvement, or is it just your expectation that this improves things in theory?

I'm concerned there's a subtle, potentially dangerous bug here.  I realize this is an already delayed response, but I'm going to hold off on making changes until I can test this thoroughly.

Most of the performance overhead is from excessive binary context switching.  For the future 0.8.x I plan to have this entire subsystem rewritten in a compiled language, with much more fine-grained use of locking.  I'm expecting this to be unnoticeable on even on low-end hardware at that time.

hllRGB:
This change is actually not available for /sbin/init. But it looks like we can achieve parallelization in a single brl enable request,I'll try that

PR #322, opened 2025-06-08
Add idlebox to brl fetch

TheOddCell:
https://github.com/TheOddCell/idlebox/

### DIFF
--- a/src/slash-bedrock/share/brl-fetch/distros/idlebox
+++ b/src/slash-bedrock/share/brl-fetch/distros/idlebox
@@ -1,0 +1,108 @@
+#!/bedrock/libexec/busybox sh
+#
+# idlebox bootstrap support
+#
+#      This is free and unencumbered software released into the public domain.
+#      Anyone is free to copy, modify, publish, use, compile, sell, or
+#      distribute this software, either in source code form or as a compiled
+#      binary, for any purpose, commercial or non-commercial, and by any means.
+#
+#      For more information, please refer to <https://unlicense.org/>
+#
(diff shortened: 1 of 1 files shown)


TheOddCell:
> Apologies for the delayed response.
It's fine :)
> In principle I'm not against the idea of generalizing `brl fetch` for more broad use cases, but I'm hesitant to do so in 0.7.x where there's a strong convention for it being specifically for something that would be an installable OS.

Makes sense.
> For 0.8.0 the plan includes having a small Bedrock-specific package manager in order to alleviate a number of consistent pain points:
> 
>     * The need for independent versioning of Bedrock components
>       
>       * This way things like `brl fetch` update churn don't drown out more notable updates
> 
>     * The need to make some components optional
>       
>       * For example, people don't need `brl tutorial` after having gone through the tutorials
> 
>     * The need to support third-party repositories
>       
>       * This exact situation

Sounds very useful!
> 
> To help with development/testing, the plan is to have some degree of backwards compatibility with 0.7.x.
> 
> Would you be open to further holding off on this MR and reworking this to be a third-party package for the package manager once it's available? Helping me test the package manager would be of quite some value, and this seems like a clean fit for a test case, as it's a drop-in file that won't interfere with the 0.7's preexisting `brl update` mechanism.
> 
> If that sounds good to you, I'll be sure to reach out once the package manager is far enough along.

Sounds good!

TheOddCell:
oops pressed wrong button, if you feel like keeping this into consideration anyways reopen it

PR #332, opened 2026-01-09
Add rebooting to init

### LINKED ISSUE #323: Reboot to init
Idea: add an command (`brl reboot`, `brl init`, or `strat -i` are some options) where you provide a stratum (and optionally, a file path) and it will reboot with that stratum selected as default, with this change being reverted to before the reboot so next (normal) reboot you are back to your default (or manually typed) init.

Example:
Let's say I have 3 stratum, `fedora`, `debian`, and `idlebox`, my own thing. Let's say I want to have a short timeout, but when I want to test `idlebox`'s init, I have to spam the number and enter till I boot else I boot with `fedora`'s init. So, after doing something in `fedora`, I can run `sudo brl reboot idlebox` and it will reboot, skip the timeout and selection, and go straight to `idlebox` as if it was the default that one time, and next reboot I would be back using `fedora`'s init as usual.

It would work like this:
`sudo brl reboot [stratum name] [/optional/path/to/init]`
or 
`sudo strat -i [stratum name] [/optional/path/to/init]` 

If no init path was selected, it would automatically detect the init binary (such as in the init selection menu). If the stratum doesn't exist or the init binary was not found, it would not reboot. If the init path was invalid or didn't work, it would just autodetect or if that doesn't work just not reboot. The no reboot part or the checking to see if it was valid could be disabled with the -F option.

TheOddCell:
`strat -i` from #323 

Resolves #323


TheOddCell:
For reasons of:
- Not tested yet
- The command itself not implemented (only added in init)

this is a draft PR.

The command not implemented is not too bad, just drop a file at `/bedrock/etc/tmpinit` with an `stratum:init` pair

